PR #2, opened 2026-02-20
Add momentum-backtest task

PQCat:
### hull-white-swaption (very_hard)

| Agent | Model | Tests passed | Reward |
|---|---|---|---|
| oracle | -- | 81/81 | 1.0 |
| claude-code | claude-opus-4-6 | 79/81 | 0.0 |
| claude-code | claude-sonnet-4-6 | timeout | 0.0 |
| claude-code | claude-haiku-4-5 | 58/81 | 0.0 |

If one tells the opus to pay attention to smoothness issue in financial engineering, then opus may pass all the tests.

### stochvol-implied-surface-new (hard)

| Agent | Model | Tests passed | Reward |
|---|---|---|---|
| oracle | -- | 64/64 | 1.0 |
| claude-code | claude-opus-4-6 | 64/64 | 1.0 |
| claude-code | claude-sonnet-4-6 | 64/64 | 1.0 |
| claude-code | claude-sonnet-4-5 | 5/64 | 0.0 |
| claude-code | claude-haiku-4-5 | 0/64 | 0.0 |

Sonnet 4-6 passes all tests but sometimes requires more than 10 write/edit attempts; we assume the rewrite number should be no more than 3.

### american-option-fd-new (hard)

| Agent | Model | Tests passed | Reward | Rewrites |
|---|---|---|---|---|
| oracle | -- | 54/54 | 1.0 | -- |
| claude-code | claude-opus-4-6 | 54/54 | 1.0 | 2–5 |
| claude-code | claude-sonnet-4-6 | 54/54 | 1.0 | 5 |
| claude-code | claude-sonnet-4-5 | 52/54 | 0.0 | 6–12 |
| claude-code | claude-haiku-4-5 | 38/54 | 0.0 | 9–11 |

If we require the agent to rewrite no more than 3 times, only opus can achieve it for sure.

### bollinger-backtest-aapl (medium)

| Agent | Model | Tests passed | Reward |
|---|---|---|---|
| oracle | -- | 35/35 | 1.0 |
| claude-code | claude-opus-4-6 | 35/35 | 1.0 |
| claude-code | claude-sonnet-4-6 | 35/35 | 1.0 |
| claude-code | claude-haiku-4-5 | 21/35 | 0.0 |

### fama-french-factor-model-new (easy) & momentum-backtest (easy)

Both passed by all agents including claude-haiku-4-5, confirming their easy difficulty rating.


foxie-huang:
This is an easy task that tests more python ability than finance. maybe qf bench needs this as floor-setting? Some improvement suggestions: 

 **Realism**

1. Add transaction costs. Include a commission_per_trade parameter (e.g., $5 or 5bps) in a params file. This is one line of extra logic but transforms the task from "can you compute EMAs" to "can you build a realistic backtest."

2. Require a buy-and-hold benchmark comparison. Add a field in results.json for buy-and-hold return over the same period. This is trivially computed but tests whether the agent understands that a strategy must be evaluated relative to a benchmark, not in isolation.

3. Add a second asset or a parameter sensitivity sweep. Either run the same strategy on two ETFs (SPY + QQQ) or run it with two parameter sets (EMA 20/50 vs EMA 50/200). This is still easy but makes the task minimally portfolio-aware.

**Distinguishing Ability**

1. Add property/consistency tests. Test that the sum of trade PnLs equals total_pnl. Test that final_capital equals initial_capital + total_pnl. Test that every sell exec_date is after its corresponding buy exec_date. These catch a much wider class of bugs than value-matching and would likely cause at least some agents to fail.

2. Remove the 4 standalone file-existence tests and add the missing pnl column check. This was already flagged in review but hasn't been addressed.

3. Add a test that the strategy underperforms buy-and-hold. Over 2000-2012, EMA crossover on SPY underperforms buy-and-hold. Testing strategy_total_return < buy_and_hold_return would verify the agent computed both correctly and understood the comparison. This is a property test that adds discriminative signal without requiring exact values.

PR #28, opened 2026-03-08
pq_cat: hull-white-swaption hints + mc-greek-surface-1 task + stochvol benchmark results

Dongzhikang:
### 1. fix(hull-white-swaption): Add OUTPUT_DIR and script-length hints to instruction.md
- Added `Critical Implementation Notes` section to help agents find output path
- Added hint about expected script length (~80–120 lines) to calibrate effort
- Oracle: ✅ 1.0 | Model evals: opus-4-6 ✅ 79/81 (97.5%), haiku/sonnet ❌ timeout

### 2. Add mc-greek-surface-1 (new task, difficulty: hard)
- Category: derivatives-pricing
- Compute European option Greeks surface via 3 methods: Finite-Difference, Pathwise, Likelihood-Ratio
- 4 option types: European call/put, Asian call/put
- 8 output files + consistency.json, 49 test cases
- Oracle: ✅ 1.0 | Model evals: opus-4-6 ✅, sonnet-4-6 ✅, sonnet-4-5 ✅, haiku-4-5 ❌ (Plan Mode)

### 3. Add benchmark results to stochvol-implied-surface-new task.toml
- Documented model evaluation results

## Oracle Status
All tasks in this PR: oracle reward = 1.0 ✅


foxie-huang:
Here are some suggestions that may help improve the task: 

**Realism**
1. Don't prescribe bump sizes. State the output units and let the agent choose. Test that FD and pathwise agree — this tests the practical judgment of bump-size selection, which is the real quant skill.
2. Add a local vol or term structure dimension. Even a two-point vol surface would make vega a partial derivative rather than a scalar sensitivity — meaningfully harder and more representative of real pricing.
3. Add a multi-strike portfolio. 3-5 options at different strikes, requiring portfolio-level Greeks aggregation and cross-strike parity checks.

**Distinguishing Ability**
1. Tighten BS comparison tolerances from 5% to 2%. With 500K paths, 5% hides systematic biases like wrong dividend handling. 2% still handles MC noise comfortably.
2. Add a convergence rate test. Verify that SE decreases as 1/√N across the path counts, catching correlated-path bugs and validating MC theory understanding.
3. Remove the explicit null enumeration from the instruction. State the principle ("pathwise requires differentiable payoffs; LR theta is unavailable") and let the agent determine which combinations are null. This shifts from instruction-following to conceptual understanding and would likely cause at least one frontier model to fail — directly addressing the saturation problem.

Dongzhikang:
# Review: PR #28 - pq_cat (hull-white-swaption, mc-greek-surface-1, stochvol-implied-surface-new)
Reviewer: Agent 🔍 | Date: 2026-04-25

## Important Note on PR Identity

The FINAL-SUMMARY lists PR #28 as "data-cleaning" with verdict 🟡 Human Review ("太简单，test 自引用"). However, the actual PR branch (`pq_cat`) contains **three entirely different tasks**: hull-white-swaption, mc-greek-surface-1, and stochvol-implied-surface-new. The prior agent review reviewed a data-cleaning task that is **not in this PR**. This review covers what's actually in the code.

---

## What This PR Does

Three substantial quantitative finance benchmark tasks:

1. **hull-white-swaption** (very_hard, interest-rate-derivatives): Calibrate Hull-White 1F to caplets, price European + Bermudan swaptions via Jamshidian decomposition and trinomial tree, compute DV01.
2. **mc-greek-surface-1** (hard, derivatives-pricing): MC Greeks (pathwise, likelihood ratio, finite difference) for European and Asian options, with BS validation, put-call parity, convergence study, delta/vega surfaces.
3. **stochvol-implied-surface-new** (hard, derivatives-pricing): Two-factor Heston (Chiarella-Ziveyi) option pricing, implied vol surface extraction, put-call parity, Dupire local vol.

---

## What I Did to Review This

- Read: task.toml, instruction.md, solve.py/solve.sh, test_outputs.py, test.sh, Dockerfile for all 3 tasks
- Verified: financial formulas, parameter admissibility, test tolerance sanity, leakage risk, difficulty calibration
- Cross-referenced: trial results (H:O:S = all pass per FINAL-SUMMARY)

---

## Scorecard

| Dimension | hull-white-swaption | mc-greek-surface-1 | stochvol-implied-surface-new |
|-----------|:---:|:---:|:---:|
| Task contract / instruction | 5/5 | 4/5 | 5/5 |
| Verifier robustness | 4/5 | 4/5 | 5/5 |
| Difficulty calibration | 3/5 | 3/5 | 3/5 |
| Model discrimination | 2/5 | 2/5 | 2/5 |
| Benchmark integrity | 4/5 | 4/5 | 4/5 |

---

## Task 1: hull-white-swaption

### Financial Formula Correctness ✅

- **HW SDE**: `dr = [θ(t) − a·r] dt + σ dW` — correct, standard Q-measure formulation.
- **ZCB pricing**: `P_HW(t,T,r) = exp(lnA - B·r)` with `B = (1-exp(-a(T-t)))/a` — correct affine model.
- **lnA formula**: Uses `lnP(T) - lnP(t) + B·f(0,t) - (σ²/4a)·B²·(1-exp(-2at))` — correct (see Brigo & Mercurio).
- **Caplet = (1+K·δ) × Put on ZCB** with `X = 1/(1+K·δ)` — correct Jamshidian trick for caplets.
- **ZCB put formula**: `σ_p = σ·B(T1,T2)·√((1-exp(-2aT1))/(2a))` — correct. `h` formula correct.
- **Jamshidian swaption decomposition**: Find r* where coupon bond = 1, then sum weighted puts — correct.
- **Trinomial tree**: `Δx = σ√(3Δt)`, `j_max = ⌈0.184/(a·Δt)⌉` — standard Hull-White tree (matches Hull textbook).
- **Arrow-Debreu calibration**: `α_n = (1/Δt)·ln(Σ Q_j·exp(-j·Δx·Δt) / P(0,t_{n+1}))` — correct.
- **Bermudan exercise**: Strictly `T_i > t_k` for remaining payments — correctly handles the common pitfall. Uses correct `α_{n_k}` per exercise date.
- **DV01**: Parallel +1bp bump, reprice on tree — correct. Uses original ATM strike (not recomputed) — correct for contract DV01.
- **Feller condition**: Not applicable (HW is Gaussian, no CIR-type positivity constraint). ✅
- **Par swap rate formula**: `(P(T0) - P(T0+τ)) / Σ δ·P(T_i)` — correct.

### Findings

#### [MINOR] Tests pin narrow numerical ranges that may be reference-solution-derived
The test file checks swaption prices against specific ranges (e.g., 1Yx5Y analytical ∈ (100, 200) bps, Bermudan ∈ (180, 320) bps). These are wide enough to be reasonable for the model, but the Bermudan premium percentage ranges (e.g., 1Yx5Y: 15-80%, 5Yx5Y: 10-60%) and the requirement that 1Yx5Y has the *largest* absolute Bermudan premium are model-specific assertions that come from running the reference solution. This is acceptable benchmark practice — they validate financial reasonableness rather than pin exact numbers.

#### [MINOR] Exercise boundary tracking has a subtle bug in solve.py
In `price_bermudan_swaption`, the exercise boundary detection at intermediate steps checks `V_new[j] == ex_val` to determine if exercise was chosen. This float equality comparison is fragile. The solve.sh version uses a cleaner `ex_val > cont_val` check. Both solutions are provided; agents won't see either. Not a blocker.

#### [NIT] Two reference solutions (solve.py and solve.sh)
Both implement the same task with slightly different structures. solve.sh uses inline Python via heredoc. This is fine but unusual — most tasks have one canonical solution.

### Assessment
Excellent task. The instruction is thorough and unambiguous — every parameter, formula, and edge case is spelled out. The Bermudan swaption component with the "strictly greater" payment filtering is a genuine difficulty trap. The trinomial tree Arrow-Debreu calibration is a non-trivial implementation. Financial formulas are all correct.

---

## Task 2: mc-greek-surface-1

### Financial Formula Correctness ✅

- **GBM under Q**: `S_T = S0·exp((r-q-σ²/2)T + σ√T·Z)` — correct with continuous dividend yield.
- **BS formulas**: d1, d2, call/put prices, all Greeks — standard and correct.
- **FD Greeks**: Central differences with specified bump sizes — correct.
- **Pathwise delta**: `E[disc · 1_{ITM} · S_T/S0]` — correct.
- **Pathwise gamma = null**: Correct — indicator function is non-differentiable.
- **LR delta score**: `Z/(S0·σ·√T)` — correct score w.r.t. log-spot.
- **LR gamma score**: `(Z²-1)/(S0²·σ²·T) - Z/(S0²·σ·√T)` — correct second derivative of log-likelihood.
- **LR vega score**: `(Z²-1)/σ - Z√T` — correct.
- **LR rho**: `payoff·(Z√T/σ - T)` — correct.
- **LR theta = null**: Correct — time score is complex.
- **Put-call parity Greeks**: `Δ_C - Δ_P = e^{-qT}`, `Γ_C - Γ_P = 0`, `ρ_C - ρ_P = KTe^{-rT}` — all correct.
- **Pathwise variance < LR variance** for delta: This is a well-known theoretical result. ✅
- **Asian call < European call**: Jensen's inequality for arithmetic average. ✅

### Findings

#### [MAJOR] FD bump size inconsistency between instruction and solution
The instruction specifies `h_delta = 0.5` (dollars), but the solution uses `h_delta = 0.005 * S0 = 0.5`. This happens to match, but the instruction says "h = 0.5" while the code computes `0.005 * S0`. If S0 were different, these would diverge. The instruction is the canonical spec, so this is fine for this specific parameterization, but the solution's approach is slightly more general. **Not a blocker** since the benchmark uses fixed S0=100.

#### [MINOR] Theta formula sign convention
The instruction says `Theta = -(V(T-h) - V(T)) / h` which equals `-(V_shorter - V_base)/h`. This gives a negative number for calls (theta decay). The solution computes `(pTd - p0) / h_theta` which is `(V(T-h) - V(T))/h` — missing the negation. However, the BS theta formula also has its own sign, so the pathwise theta and FD theta should be cross-checked. The tests don't check theta sign explicitly, only that FD matches BS within 5%. This may pass if both are consistently signed.

#### [MINOR] Asian pathwise vega derivation
The solution computes `dS_ti/dσ = S_ti·(-σ·t_i + √(dt)·cumZ_i)`. The `cumZ` term uses cumulative standard normals. This is correct: for `S_ti = S0·exp(Σ_{k=1}^{i} [(r-q-σ²/2)dt + σ√dt·Z_k])`, the derivative w.r.t. σ gives `S_ti·(-σ·t_i + √dt·Σ Z_k)`. ✅

#### [MINOR] FD Greeks reuse same seed — potential correlation
Each FD bump re-simulates with the same seed. This is **correct** and desirable (common random numbers reduce variance of difference estimators). The tests check this implicitly by requiring tight tolerances. ✅

### Assessment
Well-constructed task covering three independent MC Greek estimation methods with cross-validation. The BS benchmarks, put-call parity checks, and variance ordering provide multiple layers of verification. The instruction is clear though very long (~400 lines of implementation needed).

---

## Task 3: stochvol-implied-surface-new

### Financial Formula Correctness ✅

- **Two-factor Heston SDE**: Correct specification with `dS = μS dt + √ν1·S·dZ1 + √ν2·S·dZ2`.
- **Parameters**: `κ1 = 2^{-3/2}`, `κ2 = 2^{3/2}` — asymmetric mean-reversion speeds (slow + fast factor). `σ_m = √(κ_m·θ_m)` ensures Feller condition is exactly at boundary: `2κθ = 2σ²` → `2κθ/σ² = 2κθ/(κθ) = 2 ≥ 1`. Actually, Feller requires `2κθ ≥ σ²`, and here `σ_m² = κ_m·θ_m`, so `2κ_m·θ_m / σ_m² = 2`. Feller satisfied. ✅
- **Characteristic function**: The solution implements the standard semi-analytical form with `B`, `D1`, `D2` from the Heston-type affine model. The `Lambda`, `Theta`, `Omega`, `Q` intermediate variables follow the standard derivation. ✅
- **Integration**: Gauss-Legendre with 2000 points on [0,100] vs scipy.quad on [1e-8, 50]. The choice logic is correct.
- **Implied vol extraction**: Brentq on [1e-4, 5.0] with xtol=1e-8 — standard and correct.
- **Put pricing via characteristic function**: Uses `P_j_put = 0.5 - integral/π` (sign flip from call) — correct.
- **Dupire local vol**: `σ_loc² = (∂C/∂T + rK·∂C/∂K) / (½K²·∂²C/∂K²)` with q=0 — correct Dupire formula.
- **ρ₂₄ = -0.99**: Very high negative correlation produces strong skew. The expected IV values in tests (ATM ~0.279, skew ~0.106 at short T) are consistent with this parameterization. ✅

### Findings

#### [MINOR] Test pins very specific IV values
Tests check `IV[K=79,T=1/12] ≈ 0.334 ± 0.005`, `ATM IV ≈ 0.279 ± 0.005`, etc. These are derived from the reference solution and are specific to 4+ decimal places. This is acceptable since the CZ formula is deterministic — any correct implementation should produce the same numbers. The tight tolerances (±0.005) are appropriate for a closed-form pricing formula.

#### [MINOR] `test_chosen_method_is_gauss_legendre` assumes GL wins
The test asserts the agent must choose Gauss-Legendre. This is a reasonable expectation given the integration parameters, but it means an agent that implements both correctly but picks quad (because their quad implementation happens to be accurate enough) would fail. This is a mild form of over-specification.

#### [NIT] Solution writes to `/app/output/` (hardcoded) vs using OUTPUT_DIR
The instruction says to use `/app/output/` and the Dockerfile has WORKDIR /app. The tests also hardcode `/app/output`. This is consistent but differs from hull-white-swaption which uses the OUTPUT_DIR env var pattern. Minor inconsistency across tasks in the same PR.

### Assessment
Excellent task. The two-factor Heston model with Chiarella-Ziveyi pricing is genuinely challenging. The put-call parity verification and Dupire local vol computation add multiple verification layers. The parameter choices (ρ₂₄=-0.99) create distinctive surface shapes that are hard to fake.

---

## Cross-Task Findings

#### [MAJOR] All three models pass all three tasks — poor discrimination
Per FINAL-SUMMARY, H:O:S = all pass for this PR. For "very_hard" (hull-white-swaption) and "hard" tasks, having Haiku pass suggests:
- Either the tasks are well-specified enough that any competent LLM can implement them (which means the difficulty label is inflated), or
- The long, detailed instructions essentially serve as pseudo-code that models can transliterate.

The hull-white-swaption instruction is essentially a complete algorithm specification — it tells the agent *exactly* what formulas to use, what exercise boundary logic to implement, even warns about specific pitfalls ("CRITICAL: Use T_i > t_k strictly greater"). This level of hand-holding may undermine discrimination.

**Recommendation**: Consider removing some of the "CRITICAL" implementation hints from hull-white-swaption to increase difficulty, or relabel difficulty levels.

#### [MINOR] No answer leakage detected
Tests verify output structure and ranges, not exact pinned values (except stochvol-implied-surface-new which pins IV to ±0.005, but these are from deterministic formulas). Under Harbor, agents can't access /tests, so the pinned values aren't leaked. ✅

#### [MINOR] Consistent Dockerfile pattern
All three use `FROM finance-bench-sandbox:latest` without digest pins. Acceptable for benchmark reproducibility if the base image is controlled.

---

## Summary

**Three high-quality, financially rigorous benchmark tasks.** All formulas checked and correct. Instructions are thorough (perhaps too thorough for discrimination). The main concern is **zero model discrimination** — all three models (including Haiku) pass all three tasks, which undermines benchmark value for the "very_hard" and "hard" labels.

The prior agent review (for "data-cleaning") is **irrelevant** — it reviewed a task that isn't in this PR.

Key strengths:
- Correct financial formulas throughout (HW affine model, Jamshidian decomposition, MC Greeks, CZ two-factor Heston, Dupire)
- Multiple internal consistency checks (tree vs analytical, put-call parity, Bermudan ≥ European)
- Good test coverage with financially motivated assertions

Key concerns:
- All models pass → no discrimination signal
- Detailed instructions reduce effective difficulty
- hull-white-swaption "CRITICAL" hints are essentially debugging guides

---

## Verdict

**🟡 Human Review**

The tasks are technically excellent with correct financial content, robust verifiers, and good internal consistency checks. However, the zero discrimination (H:O:S all pass) is a significant benchmark concern — especially for a "very_hard" task. A human should decide whether:
1. The detailed instruction hints should be trimmed to increase difficulty
2. The difficulty labels should be downgraded (very_hard → hard, hard → medium)
3. The tasks are acceptable as-is for the benchmark's coverage goals despite poor discrimination

PR #38, opened 2026-03-21
feat: add zero-coupon-bootstrapping task

Dongzhikang:
test


Dongzhikang:
# Final Review: PR #38 — Zero-Coupon Yield Curve Bootstrapping

**Reviewer:** Agent 🔍 | **Date:** 2026-04-25  
**Author:** Dongzhikang  
**PR:** https://github.com/QF-Bench/QuantitativeFinance-Bench/pull/38  
**Trial Results:** H 1.0 · O 1.0 · S 1.0

---

## Verdict: ✅ Merge

Clean, financially correct task. Zero discrimination (H:O:S = 1:1:1) is the main weakness, but the task is sound as a baseline coverage item.

---

## Scorecard

| Dimension | Score | Notes |
|-----------|-------|-------|
| Task contract / instruction | 4/5 | Algorithm fully prescribed; unambiguous |
| Verifier robustness | 4/5 | Property-based tests are sound |
| Difficulty calibration | 2/5 | Labeled "hard" but all models pass trivially |
| Model discrimination | 1/5 | 1:1:1 — zero separation |
| Benchmark integrity | 3/5 | No leakage issues; tests don't pin values |
| Data realism | 3/5 | Synthetic but reasonable |

---

## Financial Correctness

**Bootstrapping methodology: ✅ Correct.**

- The par-bond pricing equation is standard: `1.0 = (c/freq) × Σ df(t_k) + (1 + c/freq) × df(T)`, solved iteratively for `df(T)`.
- Zero rate extraction via `z(T) = -ln(df(T)) / T` is correct for continuous compounding.
- The solution correctly iterates through maturities in order, using previously bootstrapped zero rates.

**Interpolation: ✅ Specified and implemented correctly.**

- Linear interpolation on zero rates for intermediate coupon dates is explicitly stated in the instruction and correctly implemented in `solve.sh`.
- Boundary handling (flat extrapolation) is reasonable.

**Forward rates: ✅ Correct.**

- Continuously compounded forward rate formula `f(T) = (z(T)·T - z(T-1)·(T-1)) / (T - T-1)` is standard.
- The solution generalizes to non-unit intervals between maturities, which is correct.

---

## Findings

### [MAJOR] Difficulty label "hard" is not justified — H:O:S = 1:1:1

**File:** `task.toml`

The task is labeled `difficulty = "hard"` with `expert_time_estimate_min = 35.0`. However, all three models (including Haiku 4.5) pass with reward 1.0. The instruction prescribes the exact algorithm step-by-step — the agent only needs to transcribe it into code. This is a code-translation task, not a quantitative reasoning task.

**Recommendation:** Relabel to `difficulty = "easy"` or `"medium"`. Not a merge blocker, but the label is misleading for benchmark consumers.

### [MINOR] Forward rate definition ambiguous for non-integer maturity gaps

**File:** `instruction.md`

The forward rate formula is defined as "1-year forward starting at T-1, for T ≥ 2" but the solution actually computes `(z(T)·T - z(T_prev)·T_prev) / (T - T_prev)` where `T_prev` is the previous maturity point, not necessarily `T-1`. If maturities include e.g. [1, 2, 3, 5, 7, 10], the "forward rate" at T=5 covers 2 years (3→5), not 1 year.

The tests don't check forward rate values directly (only presence), so this inconsistency doesn't cause failures, but the instruction is slightly misleading.

### [MINOR] No `solve.py` — only `solve.sh` with inline Python

**File:** `solution/`

The reference solution is embedded as a heredoc in `solve.sh` rather than a separate `solve.py`. This is functional but makes the oracle harder to review independently. Minor structural issue.

### [NIT] Verifier tolerance on par repricing is generous

**File:** `tests/test_outputs.py`

`test_par_bond_reprices_at_par` uses `atol=0.002`, which allows up to 20bps of pricing error. For an exact analytical bootstrapping (no MC), a tighter tolerance (e.g., `atol=1e-6`) would be more appropriate. The generous tolerance further reduces discrimination.

---

## Summary

The bootstrapping methodology is textbook-correct. Interpolation is specified (linear on zero rates). The verifier checks the right structural properties: monotonicity, DF-zero rate consistency, and par repricing. No leakage or integrity issues.

The sole material concern is **zero discrimination** — all models pass because the instruction is a complete algorithm specification. The "hard" label should be downgraded. This is acceptable as a coverage item for the pricing category but contributes no signal for model ranking.

**✅ Merge.** Optionally fix the difficulty label.

---

**⚠️ ADDENDUM (2026-04-25): Difficulty label MUST be downgraded before merge.**

H:O:S = 1:1:1 means all models pass trivially. The task is correct and should merge, but the difficulty label must be changed to **"easy"** to accurately reflect actual model performance. A benchmark with inflated difficulty labels misleads users about what the tasks actually test.

alacrity2001:
A pretty straightforward problem with clear instructions. All the equations are given and agent can just plug in and calculating. A more interesting setup would be when instruction is a bit less rigid and have more room for agent to flex around and see how robust it is. Right now might be a bit too easy for a good benchmark but also good to have in to mix the difficulty level

PR #64, opened 2026-03-26
feat: add cta-ewma-cvar task

Dongzhikang:
## Summary
CTA (Commodity Trading Advisor) strategy task that tests multi-file data handling, corporate action price adjustments, and monthly-rebalanced signal computation on 5 ETFs (SPY, QQQ, TLT, GLD, VNQ).

## Task Metadata

| Field | Value |
|-------|-------|
| **Task ID** | `cta-ewma-cvar` |
| **Difficulty** | Hard |
| **Category** | cross-domain |

## Agent Performance

| Oracle | — | 1.000 | 16/16 | — |
| claude-code | claude-haiku-4-5 | 0.000 | — | Failed on price adjustment direction |
| claude-code | claude-sonnet-4-6 | 0.000 | — | Failed on monthly observation count |
| claude-code | claude-opus-4-5 | 0.000 | 15/16 | Close — 1 test failure on max_drawdown off-by-one |

### Key difficulty axes

- **Price adjustment direction**: `factor` multiplies pre-event prices (not post-event); without adjustment, SPY shows an artificial -69% return on 2019-06-15 and TLT shows +70% on 2021-03-15
- **Monthly observation count**: `num_valid_obs` = number of monthly log-return periods (one fewer than monthly price points)
- **CF Modified VaR**: full 4-term Cornish-Fisher expansion with `bias=False`
- **Newey-West HAC t-stat**: L = floor(4·(T/100)^(2/9))

Without-skills comparison not applicable (no skills included).

## Checklist


Dongzhikang (commented):
~~This review has been superseded by the English version below.~~

Dongzhikang (commented):
## 🔍 Code Review: cta-ewma-cvar

### Findings

**[CRITICAL] Vol window description contradicts the formula**
In `instruction.md`:

> `σ_i[M] = std(r_m[M−vol_months:M−1, i], ddof=1) * sqrt(12)` (uses vol_months=3 obs from M−4 to M−2; M−1 excluded)

The formula `r_m[M−vol_months:M−1]` i.e. `r_m[M-3:M-1]` yields **2 elements** (Python slicing excludes the right endpoint). But the parenthetical says "from M−4 to M−2" which is **3 elements**.

The solution uses `vol_start = m - VOL_MONTHS - 1` (= m-4), producing 3 observations — which is the correct intent. But **an agent following the formula literally will only get 2 observations**, producing completely different results.

Fix: change the formula to `r_m[M−vol_months−1:M−1]`, or change the comment to say "M−3 to M−2".

**[MAJOR] max_drawdown computed in log space without clarification**
The solution uses `np.cumsum(r_port)` where `r_port` consists of log returns. The resulting drawdown is in log space, not percentage space. The instruction should specify which convention is expected.

**[MAJOR] hist_cvar_95 percentile method is ambiguous**
The solution uses `np.percentile(r_port, 5)` with default linear interpolation. For 57 observations, 5% = 2.85 — interpolation makes the exact cutoff ambiguous. Then `r_port[r_port <= q5]` may include 2 or 3 values. The instruction should specify the exact method (e.g., floor-based or interpolation-based).

**[MINOR] Title says "EWMA" but there is no EWMA**
`task.toml` tags include `ewma` and `ema`, but the actual signal is a simple rolling mean/std Sharpe ratio — no exponential weighting. Tags and title are misleading.

### Verdict: REQUEST CHANGES
The vol window formula contradiction is the primary issue — it will cause reasonable agent implementations to fail due to a documentation bug, not a logic error.

beckybyte:
Oracle has an off-by-one in `num_valid_obs` relative to the instruction.

`instruction.md:23` defines it as "number of monthly observations (after resampling)". After resampling in `solution/solve.py:44` there are 58 monthly price observations, but the oracle then takes log-differences at line 50 (losing one observation) and on line 115 sets:

```python
num_valid_obs = M_total  # = len(r_m) = 57
````

So the oracle reports 57, but the instruction language (consistent with "monthly observations after resampling") implies 58.

Also worth checking: line 107 uses the same `T = M_total = 57` for the Newey–West bandwidth, so the test expectation `nw_bandwidth_L = 3` was likely calibrated against the buggy count. If you fix `num_valid_obs` to 58, re-derive `L` and update the expected value in `tests/test_outputs.py:27` accordingly.

Simplest fix: either change the oracle to `num_valid_obs = len(prices_m)` (58) and re-propagate, or tighten the instruction to say "number of monthly log-returns" so that 57 is unambiguously correct.

PR #66, opened 2026-03-26
feat: add portfolio-risk-attribution task

Dongzhikang:
## Summary
Portfolio risk attribution task using a factor model approach. The agent must handle corporate-action-adjusted prices, compute factor exposures via OLS, construct a minimum-variance portfolio, decompose risk into component contributions (MCTR, percentage contribution), compute effective number of bets, and run out-of-sample evaluation with VaR/CVaR.

## Task Metadata

| Field | Value |
|-------|-------|
| **Task ID** | `portfolio-risk-attribution` |
| **Difficulty** | Hard |
| **Category** | cross-domain |

## Agent Performance

| Oracle | — | 1.000 | 24/24 | — |
| claude-code | claude-haiku-4-5 | 0.000 | 17/24 | Failed on min-var weights, MCTR, component risk, pct contribution, effective bets, OOS metrics |
| claude-code | claude-sonnet-4-6 | 0.000 | 21/24 | Close — 3 test failures |
| claude-code | claude-opus-4-5 | 0.000 | 16/24 | Failed on min_var_weight, mctr, component_risk, pct_contribution, effective_num_bets, oos_sharpe, var_99, component_var |

Without-skills comparison not applicable (no skills included).

## Checklist


Dongzhikang (commented):
~~This review has been superseded by the English version below.~~

Dongzhikang (commented):
## 🔍 Code Review: portfolio-risk-attribution

### Findings

**[MAJOR] OOS Sharpe ratio annualization is ambiguous**
The solution computes:
```python
oos_sharpe = (oos_mean_m - rf_monthly) / oos_vol_m * math.sqrt(12)
```
This equals `(ann_ret - rf_monthly×12) / ann_vol`, but `rf_monthly × 12 ≠ rf_annual` due to monthly compounding. The gap is small (~0.02 vs 0.01983), and the test tolerance of 0.05 is forgiving enough. However, the instruction should specify the exact annualization method to avoid penalizing a correct alternative implementation.

**[MAJOR] Parametric VaR ignores the mean**
The solution computes:
```python
var_99 = sigma_p * z_99
```
Standard parametric VaR is `z × σ - μ` (or `-(μ - z × σ)`). The mean is omitted here. While monthly mean is close to zero, the test tolerance is tight (0.002). An agent that includes the mean term — which is textbook correct — could fail. The instruction should either state "assume zero mean" or include the mean in the formula.

**[MAJOR] monthly_returns.csv definition is vague**
The instruction says "one row per month used in performance calculation." The solution outputs the **full-sample** portfolio returns (using min-var weights), not just OOS. An agent could reasonably interpret "performance calculation" as OOS only. The test does not validate this CSV's contents, so it doesn't affect scoring — but the spec should be precise.

**[MINOR] ENB uses only positive PCR without justification**
AAPL's PCR is negative (-0.0066) and excluded, so the remaining PCR values don't sum to 1. The entropy is therefore not over a proper probability distribution. The instruction says "use only strictly positive contributions" (correct), but a brief note explaining why would help agents understand the intent.

**[NIT] prices.csv and adjustments.csv are identical to PR #64**
Consider extracting shared data into a common directory to avoid duplication.

### What looks good
- Single-factor OLS: correct ✓
- Factor covariance matrix construction: correct ✓
- Minimum-variance portfolio (analytical solution): correct ✓
- MCTR verification (all equal at optimum — FOC holds): correct ✓
- Walk-forward OOS logic: sound ✓

### Verdict: REQUEST CHANGES
No critical blocker, but the VaR mean treatment and OOS Sharpe annualization should be made explicit in the instruction to prevent correct implementations from being penalized.

foxie-huang:
Another very sound task by the contributor, just a few places to improve: 

**Realism**

1. **Resolve the VaR mean ambiguity.** The solution computes `VaR = σ × z`, omitting the mean. The textbook formula includes it. An agent that implements the more correct version gets penalized. State "assume zero mean for parametric VaR" or include the mean — either way, make the choice explicit.

2. **Specify the exact risk-free rate conversion.** State "convert the annual risk-free rate to monthly using compound conversion: `rf_monthly = (1 + rf_annual)^(1/12) - 1`." Without this, an agent using `rf_annual / 12` produces slightly different excess returns, factor betas, and OOS Sharpe — all arguably reasonable.

3. **Clarify `monthly_returns.csv` scope.** "One row per month used in performance calculation" is ambiguous — full sample with static min-var weights, or OOS-only with walk-forward weights? The solution outputs full-sample. State this explicitly.

**Distinguishing Ability**

1. **Add a direct corporate-action adjustment test.** Check that AAPL's adjusted price on a specific pre-split date equals approximately 0.25× its raw value. Currently, adjustment errors are only caught indirectly through 20 downstream tests, making it impossible to tell whether a failure originated in data preparation or in the financial calculations.

2. **Add a test that tangency and min-var weights differ.** This property test catches agents confusing the two optimizations — tangency depends on expected returns, min-var doesn't. An agent that accidentally uses means in the min-var problem or ignores them in the tangency problem would produce identical weights for both, and this is the only test that would catch it.

3. **Expand `solution.json` intermediates.** Add one adjusted price checkpoint, the number of OOS months, and the OOS Sharpe numerator and denominator separately. The current intermediates contain only `excess_mkt_mean`, which is almost useless for diagnosing failures. Richer intermediates don't change the scoring but make the difference between "agent scored 0" and "agent failed at step 3 because of corporate action timing."

beckybyte:
The `oos_sharpe` specification is ambiguous between excess-return Sharpe and raw-return Sharpe.

`instruction.md` lines 83–85 say:

> "Report: annualized out-of-sample mean return, annualized out-of-sample volatility, and annualized out-of-sample Sharpe ratio. Annualize monthly mean by 12. Annualize monthly vol by sqrt(12)."

This never states which numerator to use for Sharpe. Meanwhile the oracle at `solution/solve.py:122` uses excess returns:

```python
oos_sharpe = (oos_mean_m - rf_monthly) / oos_vol_m * math.sqrt(12)
````

and the test expects `oos_sharpe = 1.0317 ± 0.05`.

Earlier in the instruction (`line 31`) the factor model explicitly works in excess returns, but the OOS reporting section does not carry that forward — a correct model could reasonably submit raw-return Sharpe and fail the verifier.

Also ambiguous:

* Whether `rf_monthly` is the monthly-compounded `(1 + rf_annual)^(1/12) - 1` or simple `rf_annual / 12`. The oracle uses monthly compounding — please pin this down in the instruction.
* Whether OOS returns are the simple portfolio return or the excess portfolio return.

Suggested rewording in the OOS section:

> "Compute `oos_sharpe` as `(annualized_mean_excess_return) / (annualized_volatility)`, where the monthly excess return is `portfolio_return_t - rf_monthly` and `rf_monthly = (1 + rf_annual)^(1/12) - 1`."

PR #72, opened 2026-03-26
feat: add treasury-curve-pca-butterfly task

Dongzhikang:
## Summary

Treasury yield curve analysis task covering bootstrapping, PCA decomposition, and DV01-neutral butterfly trade construction. The agent must bootstrap zero rates from par yields (semi-annual BEY → continuous compounding), compute forward rates, perform PCA on daily yield changes, calculate DV01s, construct a DV01-neutral 2s5s10s butterfly trade, and compute next-day P&L.

## Task Metadata

| Field | Value |
|-------|-------|
| **Task ID** | `treasury-curve-pca-butterfly` |
| **Difficulty** | Hard |
| **Category** | fixed-income |

## Agent Performance

| Oracle | — | 1.000 | 24/24 | — |
| Finance-Zero | GPT-4o | 0.000 | 0/24 | Script crashed on bootstrapping |
| claude-code | claude-haiku-4-5 | 0.000 | 11/24 | Bootstrapping all wrong (no iterative bootstrap) |
| claude-code | claude-sonnet-4-6 | 0.000 | 19/24 | PCA PC2 sign convention, DV01, butterfly P&L wrong |
| claude-code | claude-opus-4-5 | 0.000 | 19/24 | Same failures as Sonnet (DV01 shift convention, PCA sign, butterfly P&L) |

### Key difficulty axes

- Par yield → zero rate: semi-annual BEY → continuous compounding (requires 2·ln(1+y/2) conversion)
- Iterative bootstrapping: long-end tenors need previously solved short-end zero rates for coupon discounting
- PCA on changes vs levels: doing PCA on levels gives >95% variance in PC1 (meaningless); must use daily changes
- Forward rate: continuous compounding formula f = (z₂·t₂ - z₁·t₁)/(t₂-t₁), not simple z₂ - z₁
- DV01-neutral butterfly: weights from DV01 ratios, not naive 1:-2:1
- Cascading dependencies: bootstrap error propagates through all downstream calculations

## Checklist


foxie-huang:
Some parts to improve before merging: 

 1. Name–trade mismatch (realism). Task is called treasury-curve-pca-butterfly but builds a DV01-neutral   
  butterfly with PCA as a parallel exercise. Either rename or make the butterfly PCA-informed.

  2. PC2 sign convention is brittle. test_pca_pc2_sign_change requires ld[0] > 0 and ld[-1] < 0, but the    
  sign of a PCA eigenvector is arbitrary — both orientations describe the same factor. Reference's sign-flip
   heuristic (if np.sum(vec) < 0: flip) is unstable for PC2 because the sum is structurally near zero.
  Either loosen to "first and last loadings have opposite signs" or pin the convention explicitly in the    
  instruction ("anchor PC2 such that the shortest-tenor loading is positive").

  3. DV01 convention is implicit in the instruction. Multiple sub-conventions are unspecified: sign
  (positive = price decrease per +1bp?), one-sided vs two-sided ((P_up − P_0), (P_0 − P_down), or (P_down − 
  P_up)/2?), and "per unit notional" with bond at $100 face (divide by 100 or not?). Reference uses
  two-sided central with positive sign and /100 normalization. Pin all three explicitly:                    
       DV01 = −(P(+1bp) − P(−1bp)) / 2, divided by face value (100), reported per unit notional. Positive sign.

  4. test.sh uses pip3 instead of uvx, with silent install failure on questionable versions. Other tasks use
   uvx. Worse, pip3 install ... -q 2>&1 with no || exit swallows failures — if numpy==2.4.3, pandas==3.0.1, 
  or scipy==1.17.1 aren't installable on the target index, pytest runs against whatever's in the sandbox
  image and the test silently uses the wrong libraries. Convert to uvx and hard-fail on install error.

beckybyte:
Two correctness issues on this branch:

1. PCA sign normalization is non-deterministic.

`solution/solve.py` around lines 135–139 flips an eigenvector only when `np.sum(vec) < 0`.

This heuristic has no financial meaning and is unstable across inputs — for curve PCA the usual convention is either:

- force the first PC to have all-positive loadings
- normalize by sign of the largest-absolute loading
- align with a reference direction (e.g., long-end positive)

The current rule can silently flip in edge cases and break downstream comparisons.

2. Butterfly P&L sign/weight inconsistency.

The DV01 constraint at line 205 treats the belly as a short leg:

```python
net_dv01 = w_short * DV01_s + w_long * DV01_l - N * DV01_b
````

but the portfolio P&L at line 216 uses:

```python
mv = w_s * p_short + w_l * p_long - N_belly * p_belly
```

If `w_short`, `w_long` are meant to be notionals on the wings (long) and `N_belly` is the belly notional held short (`+1 wing / −2 belly / +1 wing` in DV01 space), the subtraction of `N_belly * p_belly` in `mv` only makes sense if `p_belly` is the full price (not a return).

Please double-check that the sign convention is consistent between the DV01 allocation step and the P&L realization step — most likely `mv` should use the return `Δp_belly` multiplied by the signed notional, not prices directly.

The expected `butterfly_pnl = 230.29 ± 50` in `tests/test_outputs.py:129` is a loose band that can hide the sign issue.

Please either:

* document the exact sign convention at the top of `solve.py` with a worked example, or
* switch to return-based P&L so signs are less error-prone.

PR #80, opened 2026-03-28
add task - credit risk - credit transition matrix and markov property (medium, credit-risk, end-to-end)

pangjacque:

- [x ]  I ran `harbor tasks check tasks/<task-id>` on my new task and ensured all checks pass
- [ ]  If external dependencies are used, versions are pinned for reproducibility

---

## Summary
Brief description of what the task tests. What skills does the agent need to demonstrate?
- The task is to build an end-to-end credit risk rating transition analysis pipeline from real S&P transition data(ESMA CEREP, 2018-2024). 
- This involves data cleaning, mapping to a standard 7+1 default scale, a cohort-weighted average transition matrix computation, multi-horizon cumulative default probabilities via matrix powers, chi-squared Markov homogeneity testing across cohorts, and deriving a continuous-time generator matrix with standard negative off-diagonal fixes.

## Task Metadata

| Field | Value |
|-------|-------|
| **Task ID** | `credit-risk-transition-matrix` |
| **Difficulty** | Medium — requires credit risk domain knowledge (rating mapping, absorbing states, censored observations), linear algebra (matrix powers, matrix logarithm), and statistical testing (chi-squared homogeneity), but follows a well-defined pipeline with no optimization or debugging traps|
| **Category** | credit-risk|
| **Skills Provided** | None — no skills folder included; agent must rely on its own domain knowledge |

## Agent Performance
- (required) run the task with the same model and scaffold but without skills. Compare with the previous run (can use metrics like number of unit tests passed). Can remove the skill by commenting out the COPY skills command in the Dockerfile or simply renaming the skills directory temporarily.
- (optional) run the task with `harbor run` with a harbor supported agent, e.g. Claude Code, Codex, Goose, etc.

|Agent| Model | Tests passed | Rewards
|-------|-------|-------|-------|
| oracle | -- | 63/63| 1.0
|Finance-Zero | - | - | 0.0(crashed) 
|claude-code | claude-opus-4-5 | 60/63 | 0.0
|codex | gpt-5.2 | 62/63 | 0.0

## Screenshots / Evidence on various agent 

- Oracle
image
image

- Finance-Zero
image

- Claude-code
    - Haiku-4-5
   
image
image

   - Opus-4-7
image
image

- Codex 
image
image

## Notes
Any additional context or explanations for checkbox items that don't fully apply.

Where agents struggle:

Metric | Oracle | GPT-5.2 (Codex) | Opus 4.7 | Haiku 4.5 | Finance Zero (Gemini 2.5 Flash)
-- | -- | -- | -- | -- | --
Score | 1.0 | 1.0 | 0.0 | 0.0 | 0.0
Tests | 107/107 | 107/107 | 104/107 (97%) | 96/107 (90%) | 0/107 (0%)
Input tokens | — | 564K | 925K | 1,452K | ~4K
Output tokens | — | 15K | 34K | 28K | ~4K
Steps | — | 24 | 28 | 40 | —
Duration | 11s | 250s | 253s | 113s | 30s

Failed Tests Root Cause
- Opus 4.7 (3 failures):

1. test_chi2_non_negative + test_p_value_bounded — wrote NaN for AAA instead of 0.0/1.0 when chi-squared test is not computable (df=0)
2. test_total_transitions_matches_cohort_sizes — off by 21
- Haiku 4.5 (11 failures):

1. 10 failures — saved CSV with ratings as index (Unnamed: 0) instead of a named rating column. Cascading KeyError across 4 test classes.
2. 1 failure — test_total_transitions_matches_cohort_sizes — off by 21
- Finance Zero / Gemini 2.5 Flash (107 failures):

1. Wrote Python inside markdown ```python ``` fences. SyntaxError on line 1, no output files produced. 8 tests failed explicitly, 99 errored on fixture loading.

**Common Pattern Analysis**
Both Opus and Haiku hit the exact same off-by-21 bug in cohort size totals — suggesting the relationship between default correction and transition counting is genuinely ambiguous in the instructions.


pangjacque:
@foxie-huang  @joyceHe703 @beckybyte  

Hi maintainers, 
Seems the Claude Code Review `action` is failing on this PR due to a **permissions** issue (the claude-review action is failing because it lacks pull-requests: write permissions for PRs from forks). This appears to be a **workflow configuration issue** on the repo side rather than anything with the PR itself.

If there's anything that needs to be enhanced or fixed, please let me know — happy to make adjustments.

Happy weekend!

Dongzhikang:
> @foxie-huang @joyceHe703 @beckybyte
> 
> Hi maintainers, Seems the Claude Code Review `action` is failing on this PR due to a **permissions** issue (the claude-review action is failing because it lacks pull-requests: write permissions for PRs from forks). This appears to be a **workflow configuration issue** on the repo side rather than anything with the PR itself.
> 
> If there's anything that needs to be enhanced or fixed, please let me know — happy to make adjustments.
> 
> Happy weekend!

Thanks for letting us know! We are aware this issue, and we will fix that!

foxie-huang:
A high quality task that may only needs some minor improvements: 

 **Realism**
1. Add a comparison to a published S&P average transition matrix. Provide a reference matrix (e.g., S&P's 1981-2023 average) and ask the agent to compute the absolute difference. The 2018-2024 window covers an unusual period (COVID defaults, post-ZIRP tightening), so differences should be meaningful. This tests whether the agent understands that its computed matrix is period-specific, not universal.

2. Add a credit VaR or expected loss computation using the cumulative PDs. Given assumed LGDs and exposures, compute the expected loss at each horizon. This connects the migration analysis to an economically useful output and tests the full credit risk chain, not just the statistical machinery.

3. Require the agent to identify which cohort year had the highest default rate and explain why. 2020 (COVID) should stand out. This tests whether the agent looks at the data rather than just running formulas — a basic but important analytical skill.

**Distinguishing Ability**

1. Add at least one reference value test. Test a specific transition probability (e.g., BBB→Default 1-year average ≈ X%) against the oracle value. The purely invariant-based approach is robust but can't catch data-cleaning errors that produce structurally valid but numerically wrong matrices. One or two reference checks would close this gap.

2. Add a generator matrix round-trip test. Test that exp(Q) approximates P̄ within a reasonable tolerance (e.g., max absolute difference < 0.01). This validates the entire generator pipeline — logarithm, negative-fix, row-sum adjustment — as a self-consistent system. Currently, the generator tests only check structural properties (row sums zero, off-diagonals non-negative) but not numerical correctness.

3. Test that the chi-squared sparse-column filtering was applied. Add a test checking that the degrees of freedom for AAA's chi-squared test is less than (n_cohorts - 1) × (n_destinations - 1) — since AAA has very few observations, several destination columns should be filtered out, reducing the degrees of freedom. This catches agents that skip the filtering step.

PQCat:
claude-sonnet-4-6 aced it: PERFECT — 63/63 passed ✅

63/63 tests passed
Only 8/30 turns used (finished very efficiently)
83 seconds, 114K input / 6K output tokens

PQCat:
Haiku 4.5 also aced it, though it took more effort:
        Model | Tests | Turns | Time | Tokens (in/out)
Sonnet 4.6 | 63/63 ✅ | 8/30 | 83s | 115K / 6K
Haiku 4.5 | 63/63 ✅ | 28/30 | 165s | 697K / 19K

pangjacque:
Hi @foxie-huang ,
Thank you for the detailed feedback! I really appreciate the suggestions. I’ve **addressed most of the items** in the latest several commits; please let me know if any further adjustments are needed."
> 
> **Realism**
> 
> 1. Add a comparison to a published S&P average transition matrix. Provide a reference matrix (e.g., S&P's 1981-2023 average) and ask the agent to compute the absolute difference. The 2018-2024 window covers an unusual period (COVID defaults, post-ZIRP tightening), so differences should be meaningful. This tests whether the agent understands that its computed matrix is period-specific, not universal.

Addressed. I adde the S&P 1981-2024 one-year average transition matrix ( source : table 33 from "2024 Annual Global Corporate Default And Rating Transition Study" as `reference_transition_matrix.csv` in data. The output `transition_matrix_deviation.csv` would capture per-cell deviations and reported in `summary.json`.   We confirmed that the largest deviation is CCC-> Default with ~10pp. 

> 2. Add a credit VaR or expected loss computation using the cumulative PDs. Given assumed LGDs and exposures, compute the expected loss at each horizon. This connects the migration analysis to an economically useful output and tests the full credit risk chain, not just the statistical machinery.

This is the only item not addressed. This task is actually substantially covered by my sibling tasks [PR](https://github.com/QF-Bench/QuantitativeFinance-Bench/pull/117), which computes EL, VaR, and CVaR using a real corporate bond portfolio with Monte Carlo simulation, Gaussian and Student-t copulas, and Euler allocation. Adding a simplified version here would be redundant.

> 3. Require the agent to identify which cohort year had the highest default rate and explain why. 2020 (COVID) should stand out. This tests whether the agent looks at the data rather than just running formulas — a basic but important analytical skill.

Addressed.  Actually this one is with a data correction that I discovered that the CEREP transition matrix undercounts defaults because issuers whose rating was withdrawn before defaulting appear in the Withdrawals column, not the Default column. I added `Default-Rates-*.csv` files downloaded from the CEREP Default Rates tab (which captures all default events regardless of withdrawal status) and a correction step that moves hidden defaults from Withdrawals into the Default column. With this correction, 2020 (COVID) correctly emerges as the peak default with ~ 3.85%.

> **Distinguishing Ability**
> 
> 1. Add at least one reference value test. Test a specific transition probability (e.g., BBB→Default 1-year average ≈ X%) against the oracle value. The purely invariant-based approach is robust but can't catch data-cleaning errors that produce structurally valid but numerically wrong matrices. One or two reference checks would close this gap.

Addressed. Added 5 oracle value tests in 1TestReferenceValues1. The CCC→Default test is the most valuable — without the default-rate correction it would be ~3% instead of ~31%, but still pass structural invariant test. 

> 2. Add a generator matrix round-trip test. Test that exp(Q) approximates P̄ within a reasonable tolerance (e.g., max absolute difference < 0.01). This validates the entire generator pipeline — logarithm, negative-fix, row-sum adjustment — as a self-consistent system. Currently, the generator tests only check structural properties (row sums zero, off-diagonals non-negative) but not numerical correctness.

Addressed. Added `test_generator_round_trip` in `TestCrossFileConsistency` which loads both `average_transition_matrix.csv` (P̄) and `generator_matrix.csv` (Q), computes exp(Q) via `scipy.linalg.expm`, and asserts max|P̄ - exp(Q)| < 0.01.  This validates the entire generator pipeline — matrix logarithm, negative off-diagonal fix, and row-sum adjustment — as a self-consistent system.


> 3. Test that the chi-squared sparse-column filtering was applied. Add a test checking that the degrees of freedom for AAA's chi-squared test is less than (n_cohorts - 1) × (n_destinations - 1) — since AAA has very few observations, several destination columns should be filtered out, reducing the degrees of freedom. This catches agents that skip the filtering step.

Addressed. Added two tests: `test_sparse_column_filtering_aaa` checks that AAA's degrees of freedom is less than the unfiltered maximum of (7-1)×(8-1) = 42 (actually  dof = 0, since AAA has only ~2 issuers per cohort and nearly all destination columns have < 5 total observations);  `test_sparse_column_filtering_reduces_df` checks that AAA df < BBB df, since sparser ratings should have more columns filtered out.

pangjacque:
> Haiku 4.5 also aced it, though it took more effort: Model | Tests | Turns | Time | Tokens (in/out) Sonnet 4.6 | 63/63 ✅ | 8/30 | 83s | 115K / 6K Haiku 4.5 | 63/63 ✅ | 28/30 | 165s | 697K / 19K

Hi @PQCat,

Thanks for the tests! I've addressed the improvements suggested by @foxie-huang and updated the results in the  writeup <Note> section. The latest changes seem to have challenged the Haiku model quite a bit. Appreciate the support!

pangjacque:
Hi @Dongzhikang @PQCat , 

The PR updated triggers Claude Code Review, which is constantly failed. I checked the error message above, and it seems a GitHub Actions CI configuration issue. The Claude Code Review action **can't authenticate because the API key is missing or expired**. 
Just want to comment here to let you guys (or whoever manages the repo secretes) know :)

alacrity2001:
This one feels a bit conventions-heavy, and the main trick seems to be around the convention rather than the pipeline itself. In particular, the requirement that `total_transitions == sum(cohort_sizes)` effectively forces you to exclude Default-starting cohorts, but that’s not stated anywhere in the instruction. A reasonable implementation could sum the full matrix and fail by ~20, which is what we’re seeing.

Also the solution works partly because of how variables are handled. The variable `total_C` gets rebound to the non-default cohort totals in line 190 and line 454.

`return P_avg, total_C[:N_NON_DEFAULT].sum(axis=1)`

`P_avg, total_C = step3_average_matrix(cohort_counts)` 

So you end up matching the test without explicitly stating the convention. The variable rebinding is a bit confusing there, but it does produce the expected output.

I think excluding default cohorts is defensible in credit transition analysis. But if that’s the intended convention, I  suggest making it explicit in the instruction (e.g., define `total_transitions` as the sum over non-default starting cohorts). Otherwise, the task ends up testing whether the model guesses the convention rather than executes the logic.

If your goal is to test this specific domain intuition, then this is probably fine. If not, I would lean toward making that piece explicit.

pangjacque:
> This one feels a bit conventions-heavy, and the main trick seems to be around the convention rather than the pipeline itself. In particular, the requirement that `total_transitions == sum(cohort_sizes)` effectively forces you to exclude Default-starting cohorts, but that’s not stated anywhere in the instruction. A reasonable implementation could sum the full matrix and fail by ~20, which is what we’re seeing.
> 
> Also the solution works partly because of how variables are handled. The variable `total_C` gets rebound to the non-default cohort totals in line 190 and line 454.
> 
> `return P_avg, total_C[:N_NON_DEFAULT].sum(axis=1)`
> 
> `P_avg, total_C = step3_average_matrix(cohort_counts)`
> 
> So you end up matching the test without explicitly stating the convention. The variable rebinding is a bit confusing there, but it does produce the expected output.
> 
> I think excluding default cohorts is defensible in credit transition analysis. But if that’s the intended convention, I suggest making it explicit in the instruction (e.g., define `total_transitions` as the sum over non-default starting cohorts). Otherwise, the task ends up testing whether the model guesses the convention rather than executes the logic.
> 
> If your goal is to test this specific domain intuition, then this is probably fine. If not, I would lean toward making that piece explicit.

Hi @alacrity2001 ,

Thanks for the valuable feedback!

Yes, the decision to leave the treatment of Default-starting cohorts implicit is **an intentional design choice**. If every formula, edge case, and data constraint were explicitly detailed in the `instruction.md`, the task would be reduced to a simple translation of text into code, however, in real-world credit infrastructure requirements are often under-specified. 

My goal for this task is to ensure the agent is 'thinking' in terms of the credit risk lifecycle, not just 'executing' .

PR #87, opened 2026-04-02
feat: add yield-curve-bootstrap-immunization task (hard)

harvenstar:
**If your PR is adding a new task to QuantitativeFinance-Bench, please complete this by adding an \"x\" next to each applicable item.**

- [x]  I ran `harbor tasks check tasks/yield-curve-bootstrap-immunization` on my new task and ensured all checks pass

---

## Summary

Full fixed-income pipeline using real FRED Treasury par yield data (2024-01-02). The agent bootstraps zero-coupon discount factors from par yields, computes spot/forward rates, prices four T-notes with full analytics (YTM, Macaulay/modified duration, convexity, key rate durations, z-spread), solves a liability immunization via SLSQP (PV + KRD_2 + KRD_5 match), runs three stress scenarios (parallel +100bp, steepener, butterfly), and fits a Svensson (1994) six-parameter term-structure model to the bootstrapped spot curve.

## Task Metadata

| Field | Value |
|-------|-------|
| **Task ID** | `yield-curve-bootstrap-immunization` |
| **Difficulty** | Hard — pinned tests on Svensson β₀ and λ₂ discriminate against Nelson-Siegel (4-param) implementations and against Svensson fits with wrong starting values. The full fixed-income pipeline (bootstrap → KRD → immunization → stress → Svensson) requires consistent handling of conventions across 6 output files and 52 tests. |
| **Category** | fixed-income |
| **Skills Provided** | No skills provided |

## Oracle

Verified passing locally — reward=1.0, 52/52 tests.

```
PV_liability = 658,385.35
Portfolio KRDs: 2Y=0.2233, 5Y=1.2535, 10Y=0.0000
Liability KRDs: 2Y=0.2233, 5Y=1.2535, 10Y=0.0000 (matched)
```

Full logs: https://gist.github.com/harvenstar/5a53e0b3b9ddcf98ccec4602b30fc13d

## Agent Performance

The most recent deploy server runs showed oracle crashes on the server side that I could not reproduce on my local machine (both ARM and x86 Docker pass cleanly). Collaborators investigating server-side logs reported `rewardfilenotfound`, which suggests the test harness exited before writing `/logs/verifier/reward.txt`. Because none of those runs produced valid agent test counts (all showed `tests=—`), there are no trustworthy deploy-server agent results to paste yet — I have asked @joyceHe703 / @Dongzhikang to re-trigger `/test-deploy` on the latest commit `86728be`.

Expected agent failure modes based on the task design:

#### finance-zero + gemini-3-flash-preview
Expected to fail the whole pipeline. finance-zero is a minimal agent without the multi-step reasoning needed to chain bootstrap → analytics → KRD → SLSQP → stress → Svensson, so most output files will be missing or malformed.

#### claude-code + claude-opus-4-6 / claude-haiku-4-5
Expected to fail the pinned Svensson tests (`test_ns_beta0_pinned`, `test_ns_lambda2_pinned`). The Svensson SSE surface is non-convex, so agents that implement the 4-parameter Nelson-Siegel by mistake — or the 6-parameter Svensson with a different initialization — converge to a different local optimum and produce β₀/λ₂ outside the pinned tolerance. May also trip on `test_ba_z_spread_*` if the agent skips the z-spread step, and on `test_ip_krd*_match` if the SLSQP constraints are wired up wrong.

The hardest discriminators are the two Svensson pinned tests: they catch both (a) fitting the simpler NS model and (b) Svensson with incorrect starting values. Agents that satisfy every other test can still fail these.


harvenstar:
image

harvenstar:
Hi @Dongzhikang, the oracle failure on the deploy server seems to be a transient infrastructure issue — all agents including oracle returned tests=0/0 which indicates the verifier didn't run at all rather than a code problem. Verified the oracle passes locally 3 times in a row (reward=1.0). Could you re-trigger `/test-deploy` when you get a chance? Thanks!

harvenstar:
Hi @Dongzhikang, the oracle failure on commit `dbeddc4` was caused by two issues:

1. **scipy not pinned** in Dockerfile (`scipy>=1.10`) — a newer scipy version changed SLSQP convergence behavior, causing the immunization optimization to fail and crash the oracle with a `RuntimeError`
2. **numpy/pandas also not pinned** in test.sh

Fixed in commit `fb33f23`:
- Dockerfile now pins `scipy==1.15.2`
- solve.sh adds a retry with relaxed tolerance if tight convergence fails (no more RuntimeError crash)
- test.sh pins `numpy==2.2.3`, `pandas==2.2.3`, `scipy==1.15.2`

Could you please re-trigger `/test-deploy`? Thanks!

harvenstar:
Updated in commit `d435589`:

The oracle failure was due to the Svensson fit being **non-convex** — different hardware converges to different local optima, causing the pinned tests to fail. Fixes:

1. **instruction.md**: Added explicit starting values `x0 = [r(10), r(1)−r(10), 0.5, −0.5, 1.0, 5.0]` — this also resolves the `harbor tasks check` 'Behavior In Task Description' failure
2. **solve.sh**: Multi-start optimization (3 starting points, keeps best result) for cross-platform robustness

@joyceHe703 @Dongzhikang Could reviewer please re-trigger `/test-deploy`? Thanks so much!

harvenstar:
Local oracle run on commit `d435589` — **reward=1.0**, all tests pass:

```
        oracle on tasks         
┏━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓
┃ Metric              ┃ Value  ┃
┡━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩
│ Agent               │ oracle │
│ Dataset             │ tasks  │
│ Trials              │ 1      │
│ Errors              │ 0      │
│                     │        │
│ Mean                │ 1.000  │
│ Mean_Reward         │ 1.000  │
│ Mean_Reward_Count   │ 1      │
│                     │        │
│ Reward Distribution │        │
│   reward = 1.0      │ 1      │
└─────────────────────┴────────┘
```

Built `finance-bench-sandbox:latest` locally from `docker/sandbox.Dockerfile` and ran via `harbor run`. Previous deploy server failures were due to the Svensson fit converging to a different local optimum — fixed with multi-start optimization and explicit starting values in the instruction.

Ready for `/test-deploy` re-trigger.

harvenstar:
@joyceHe703 @Dongzhikang I've verified the oracle passes on both ARM and x86 (Docker `--platform linux/amd64`) locally — 52/52 tests pass on both.

The CI workflow log only shows `Oracle reward: 0.0` without the actual error. Could someone with deploy server access check the harbor job logs for `td-yield-curve-bootstrap-immunization-8de5e4`? The detailed error should be in the harbor job directory on the server.

The code itself is correct — this seems to be an environment-specific issue on the deploy server.

harvenstar:
Hi @joyceHe703 @Dongzhikang, a few updates:

1. Fixed `author_email` in `task.toml` (was empty, now set to `harvenx01@gmail.com`) — commit `cf49d8a`
2. Local oracle verified passing on both ARM and x86 (`docker --platform linux/amd64`): **52/52 tests pass, reward=1.0**

Since the deploy server keeps showing `Oracle reward: 0.0` but the CI logs only capture the final reward without details, could someone with server access check the harbor job logs (e.g. `jobs/td-yield-curve-bootstrap-immunization-8de5e4/`) to see why the oracle fails on the server? That would help a lot in debugging. Thanks!

harvenstar:
Thanks for checking! The `rewardfilenotfound` error means `/logs/verifier/reward.txt` was never written — which suggests `test.sh` didn't execute or crashed before completing.

Could you share the full harbor output when running the oracle? Something like:

```bash
harbor run --path ./tasks --task-name yield-curve-bootstrap-immunization -a oracle 2>&1
```

Specifically the job log (e.g. `jobs/<job-name>/job.log`) would show whether `solve.sh` or `test.sh` is the one failing. On my end (both ARM and x86 Docker), everything passes — 52/52 tests, reward=1.0.

harvenstar:
Good catch — the UV install might be failing in the container. Updated `test.sh` in commit `8760951`: replaced `uvx` with direct `pip install` so it doesn't depend on downloading UV at runtime. Could you try again?

harvenstar:
Thanks for trying! The error details should be in the harbor trial directory. After running:

```bash
harbor run --path ./tasks --task-name yield-curve-bootstrap-immunization -a oracle
```

Check the trial output in the jobs directory:
- `jobs/<job-name>/<trial-folder>/verifier/test-stdout.txt` — **this will show exactly why test.sh failed** (pip errors, pytest errors, etc.)
- `jobs/<job-name>/<trial-folder>/agent/oracle.txt` — solve.sh output

Could you paste the contents of `test-stdout.txt`? That should reveal the root cause of the `rewardfilenotfound` issue.

joyceHe703:
@harvenstar Hi contributor, you should be able to access harbor job logs? 
Screenshot 2026-04-15 at 12 18 14 AM

harvenstar:
@joyceHe703 Yes, I have my local harbor job logs — and they all pass. Here's the output:

**agent/oracle.txt:**
```
Complete. PV_liability=658385.35
Port KRDs: 2Y=0.2233 5Y=1.2535 10Y=0.0000
Liab KRDs: 2Y=0.2233 5Y=1.2535 10Y=-0.0000
```

**verifier/test-stdout.txt:**
```
downloading uv 0.9.5 aarch64-unknown-linux-gnu
...installing to /root/.local/bin
...
============================== 52 passed in 0.61s ==============================
```

**verifier/reward.txt:** `1`

Everything passes on my end (52/52 tests, reward=1.0). The issue only reproduces on the deploy server and your local machine.


May I ask the error message in your local `verifier/test-stdout.txt`, which shows exactly what failed.  Thanks so much!!

harvenstar:
@joyceHe703 Here are my full local harbor job logs: https://gist.github.com/harvenstar/5a53e0b3b9ddcf98ccec4602b30fc13d

- `oracle.txt`: solve.sh completes successfully
- `test-stdout.txt`: UV installs, pytest runs 52/52 pass
- `reward.txt`: `1`

joyceHe703:
@harvenstar , I see, the oracle also passed for me. Let me re-trigger the test-deploy. If it's still not passing oracle, I will ask my teammate for logs.

Dongzhikang:
> @harvenstar , I see, the oracle also passed for me. Let me re-trigger the test-deploy. If it's still not passing oracle, I will ask my teammate for logs.

@harvenstar @joyceHe703 
I can run oracle on my laptop as well, so I think oracle should be fine

Jingyi-Jia (changes requested):
Overall good but requesting minor changes:

Economically, this is a reasonable fixed-income task: bootstrap a Treasury curve, compute bond analytics and KRDs, immunize a liability, and fit Svensson. 

## What Looks Good

- Real fixed-income workflow rather than a toy math exercise.
- Good coverage of term-structure, duration, key-rate, and immunization ideas.
- Svensson fitting is a legitimate source of difficulty if the convention is pinned cleanly.

## Key Suggestions

- The instruction says the optimizer must match PV, KRD_2, and KRD_5, but the verifier also appears to require KRD_10. Can you double check? The immunization target needs to be explicit. If KRD_10 is part of the required hedge, say so. If it is not, then we should not silently test it.
- The instruction says to raise if SLSQP fails, while later oracle behavior appears to fall back to a best iterate. Can you check again? Thanks! 
- The Svensson fit should be hard because term-structure estimation is genuinely challenging, not because the optimizer is fragile across environments. And if Svensson starting values are benchmark-critical, state them clearly and justify why.

Thanks a lot!

harvenstar:
Thanks @Jingyi-Jia! Pushed 86728be addressing all three points:

On KRD_10 — you're right, the immunization only targets PV + KRD_2 + KRD_5. KRD_10 just happens to land at zero because the liability has no cash flows beyond year 7 (so the hedge automatically matches the liability's zero KRD_10). Updated the instruction to say KRD_10 is reported as a diagnostic, not a target.

On SLSQP convergence — fair point. The oracle retries once at relaxed ftol (added back when I was chasing cross-platform numerical issues), so I spelled that out in the instruction instead of leaving `raise RuntimeError` stale.

On Svensson starting values — added a brief note explaining why the x0 is pinned. Non-convex SSE surface + L-BFGS-B means different environments converge to different local optima without a fixed start; Diebold & Rudebusch (2013) recommend the β₀=long, β₁=short−long initialization anyway.

joyceHe703:
@harvenstar Thanks for addressing the comments, have updated the PR description with the new model test results? 

Thank you!

harvenstar:
@joyceHe703 Thanks for asking! Already updated the PR description with the new model test results to reflect the current state.

PR #91, opened 2026-04-03
feat: add polars-api-migration (medium)

owen8877:

---

## Summary
This task tests a developer's proficiency in navigating breaking API changes and library deprecations during a major version transition in Polars. It specifically evaluates the ability to identify and update core data manipulation patterns, such as renaming `groupby` to `group_by`, adjusting sorting parameters from reverse to descending, and transitioning from apply to more explicit mapping methods. The task is categorized as medium because it demands attention to nuanced shifts, like the move to horizontal logical expressions and specific list-processing methods, which are essential for maintaining code correctness and performance in modern Polars.

## Task Metadata

| Field | Value |
|-------|-------|
| **Task ID** | `polars-api-migration` |
| **Difficulty** | Medium- The agent needs to perform api migration while keeping accuracy and efficiency |
| **Category** | debug-migration |

## Agent Performance

### Oracle Run
oracle
oracle test

#### Opus 4.6
opus

- time used: 09:21

#### Sonnet 4.6
sonnet
sonnet test

- time used: 06:03
- failure analysis:
  + case 17: does not handle null values when joining on a column that has nulls (see below; first is the desired behavior and second is sonnet output behavior, missing the key `nulls_equal=True` argument)
null-ticker-new
null-ticker-sonnet

#### Haiku 4.5
haiku test

- time used: 04:21
- failure analysis:
  + case 17: same as sonnet
  + case 13: WMT and UPS are not included; this is due to an api change in `is_first`

#### Finance Zero (gemini 2.5 flash lite)
zero
zero test
zero-agent-output

- time used: 00:27
- failure analysis: does not produce a working python script, instead directly outputs transformed result on the pilot set


PQCat:
Haiku4.5: 21/24 passed — 3 failing. 3 issues remain:

1. test_polars_version_content — Installed Polars on this machine is 1.36.1 but the reference expects 1.39.3. This is an environment mismatch, not a code issue — the task was authored on a machine with Polars 1.39.3.
2. test_log_PolarsInefficientMapWarning — There's still a map_elements call somewhere triggering this warning. The agent thought it eliminated it but the log still contains it (likely being piped from pytest output itself into the log file).
3. test_log_DeprecationWarning — Same issue — the pytest run output is being captured into the log, bringing in warnings from other code paths.

The version mismatch (#1) is a local environment issue — not something the agent can fix. Tests 2 and 3 are about the log file capturing too much (including pytest's own output). All 10 step outputs are numerically correct. This task likely requires Polars 1.39.3 to pass cleanly in Docker.

PQCat:
The last two failing log tests are actually environment/setup issues, not instruction issues:

1.The instruction already correctly says "avoid anti-patterns" — it just doesn't explicitly name map_elements with return_dtype or PolarsInefficientMapWarning. But in Docker, the log would only contain warnings from the agent's own code, making this a valid test.

2.The instruction does need one clarification though — it should explicitly state:
"Your migrated code must run without any PolarsInefficientMapWarning or DeprecationWarning. These warnings indicate anti-patterns — use vectorized expressions instead of map_elements, use pl.len() instead of pl.count(), etc."

Currently, the instruction only says "use best practice that avoid anti-patterns" without naming what those anti-patterns are. Adding this explicit constraint would make it clear to the agent what to avoid, rather than discovering it only from the log test failures.

PQCat:
After modify the task based on above suggestions: 24/24 passed on the original test suite. PR #91 (polars-api-migration) with the modified instruction is a full PASS — all file existence, content validation, and zero-warning checks confirmed.

To summarize the round 3 results for PR #91:

Haiku 4.5 with the modified instruction (polars==1.39.3 explicit, anti-pattern constraints added)
24/24 tests passed including test_log_PolarsInefficientMapWarning and test_log_DeprecationWarning
Log file is clean — no warnings in pipeline output

Hi @owen8877, thanks for your task, could you please follow the above suggestions to update the task accordingly?

owen8877:
@PQCat Thanks for the investigation. I have updated the `instruction.md` accordingly and ran a local test using Haiku - it passed the tests.

Changes: https://github.com/owen8877/QuantitativeFinance-Bench/commit/cf96158f3ddf2210b32097542ef32f1138adebba

Could you check this again and approve the merge? Much appreciated.

Dongzhikang:
## Review: PR #91 — polars-api-migration
Reviewer: Grim 🔍 | Date: 2026-04-16

- URL: https://github.com/QF-Bench/QuantitativeFinance-Bench/pull/91
- Status: local draft updated to align with Harbor runtime review standard

### What This PR Does (Plain Language)
This PR adds a benchmark task where an agent must migrate a script from one Polars API style to another and preserve the expected staged outputs. In plain terms, the agent is performing a focused code-migration exercise rather than a deep financial-analysis task.

Review focus: the main questions here are difficulty calibration and benchmark value. A deterministic migration task can be valid, but the claimed difficulty has to match what the task is actually asking the agent to do.

### What I Did to Review This
- Reviewed the existing local draft and updated it to the current Harbor runtime review standard.
- Rechecked the earlier conclusions with focus on task difficulty, anti-overfitting posture, and grading reliability.
- Preserved the core judgment that this looks much easier than the claimed level.

### Scorecard

| Dimension | Score | Weight | Weighted |
| --- | ---: | ---: | ---: |
| A. Structural Completeness | 7 / 10 | 20% | 1.4 |
| B. Task Contract & Determinism | 8 / 10 | 20% | 1.6 |
| C. Difficulty Calibration & Model Signal | 2 / 10 | 35% | 0.7 |
| D. Evaluation Criteria & Grading Reliability | 4 / 10 | 25% | 1.0 |
| Total |  |  | 4.7 / 10 |

Threshold: 7.0 → 🔴 REQUEST CHANGES

### A. Structural Completeness — 7/10
The task is concrete and deterministic. As a migration exercise, it is easy to understand.

### B. Task Contract & Determinism — 8/10
The task goal is narrow and specific, which makes it straightforward to grade reproducibly.

### C. Difficulty Calibration & Model Signal — 2/10
**[MAJOR] This does not look like a meaningful medium benchmark task**  
The prior review notes indicate the API mapping is explicit, the required output is narrow, and the verifier compares against fixed known stage outputs. That looks much closer to an easy migration exercise than a meaningful medium benchmark item.

Suggested fix: downgrade the difficulty label unless the author can provide strong contrary evidence.

**[MAJOR] Hardcoding / overfitting resistance looks weak**  
The task shape, dataset, staged outputs, and reference artifacts appear fixed enough that an agent can optimize toward them fairly directly.

Suggested fix: strengthen the anti-overfitting posture if the task is meant to carry benchmark weight beyond a simple migration check.

### D. Evaluation Criteria & Grading Reliability — 4/10
**[MAJOR] Verifier failure handling is reportedly still brittle**  
The prior review notes indicate `tests/test.sh` uses the fragile `$?` pattern after the main command rather than an explicit `if ...; then ... else ... fi` wrapper.

Suggested fix: rewrite the shell wrapper in explicit form so reward emission is reliable.

### Summary
This task may be useful as a deterministic migration exercise, but it does not currently look like a credible medium benchmark item. The main problem is not ambiguity, it is that the claimed challenge level appears too high for the actual work required.

| Priority | Action |
| --- | --- |
| 🔴 Must fix | Recalibrate difficulty, likely below medium |
| 🔴 Must fix | Fix verifier reward handling |
| 🟠 Should fix | Strengthen anti-overfitting design if this is meant to be a benchmark-quality task |

### Verdict
REQUEST CHANGES

The task is clear, but the current benchmark positioning does not look well calibrated.

---
Posted from updated Harbor-aligned reviewer workflow on 2026-04-16.

owen8877:
@Dongzhikang Thanks for reviewing the task. Yet, I don't think the logic of this review is strongly supported and this is a task that matches the "medium" difficulty, for the following reasons.

> The prior review notes indicate the API mapping is explicit, the required output is narrow, and the verifier compares against fixed known stage outputs. That looks much closer to an easy migration exercise than a meaningful medium benchmark item.

This review does not provide material evidence on why the API mapping is "explicit". Quite on the contrary, I have handcrafted these exercises by observing the patterns from a recent polars 0->1 migration from my own experiences. The nuisances in which API should be replaced is not straight-forward; in fact, there is no such guidance even in the polars official document. Anyone (or agent) working on this task would not claim this as an "easy migration", since the target persona defined in the [task contribution document](https://github.com/QF-Bench/QuantitativeFinance-Bench/blob/main/docs/task_contribution.md) mentions an easy task should only involve "**basic** API calls", rather than dealing with debugging issues (as is involved in the hard difficulty).

> The task shape, dataset, staged outputs, and reference artifacts appear fixed enough that an agent can optimize toward them fairly directly.

This argument does not stand on a solid ground. Even if the agent could run the given script under the old api, it still needs to resolve version conflicts (which is a tricky problem to deal with), let alone switching between and comparing the outputs. So the agent has to observe the pattern within the code and adapt accordingly, rather than optimizing toward a "fixed" target. One important notice is that only the script under the old api is given, but not the artifact it produces, thus it is already a non-trivial task to obtain the "ground truth" artifact to begin with.

> The prior review notes indicate tests/test.sh uses the fragile $? pattern after the main command rather than an explicit if ...; then ... else ... fi wrapper.

This is a common pattern in the merged tasks (see example [1](https://github.com/QF-Bench/QuantitativeFinance-Bench/blob/main/tasks/momentum-backtest/tests/test.sh), [2](https://github.com/QF-Bench/QuantitativeFinance-Bench/blob/main/tasks/mc-greeks-surface/tests/test.sh), [3](https://github.com/QF-Bench/QuantitativeFinance-Bench/blob/main/tasks/mc-greek-surface-1/tests/test.sh). and [4](https://github.com/QF-Bench/QuantitativeFinance-Bench/blob/main/tasks/american-option-fd-new/tests/test.sh)) and I would put it on the lowest priority unless there is a unified change to all the test cases.

To conclude, I think the current status of this test case lives up to the standard of a medium test and I would request the reviewer to look more closely at the details of the test rather than making unfounded, wide-scoped claims.

joyceHe703 (approved):
LGTM, thanks for updating the test cases and the PR description for the updated model test results!

PR #114, opened 2026-04-06
feat: Add brinson-sector-attribution task (Medium)

boqiny:

---

## Summary
Brief description of what the task tests. What skills does the agent need to demonstrate?

This task tests whether an agent can implement Brinson-Fachler performance attribution on real sector ETF data, decomposing a portfolio's active return into allocation, selection, and interaction effects. The key challenges are getting the BF formula right, properly handling a cash allocation that doesn't exist in the benchmark, and tracking portfolio weight drift between quarterly rebalancing dates. 

## Task Metadata
| Field | Value |
|-------|-------|
| **Task ID** | `brinson-sector-attribution` |
| **Difficulty** | Medium |
| **Category** | performance-attribution, data-processing |
| **Skills Provided** | - |
| **Data Source of Input Files** | Yahoo Finance |

## Agent Performance
- (required) run the task with the same model and scaffold but without skills. Compare with the previous run (can use metrics like number of unit tests passed). Can remove the skill by commenting out the COPY skills command in the Dockerfile or simply renaming the skills directory temporarily.
- (optional) run the task with `harbor run` with a harbor supported agent, e.g. Claude Code, Codex, Goose, etc.

### Harbor tasks check
Screenshot or output of harbor tasks check `harbor tasks check tasks/<task-id>` passing
image

### Oracle Run
Screenshot showing oracle passing all tests
image

### Agent Runs
Screenshots of agent runs with and without skills. 
Summarize and explain all unit test failures for each agent run.
Also note here the `final_metrics` in `jobs/date/task-name__xxxxxx/agent/trajectory.json` for each run.
Finance-Zero:
image
image

Codex - gpt5.2 (updated):
image
```
"final_metrics": {
    "total_prompt_tokens": 259369,
    "total_completion_tokens": 8232,
    "total_cached_tokens": 217216,
    "total_steps": 17,
    "extra": {
      "reasoning_output_tokens": 4273,
      "total_tokens": 267601,
      "last_token_usage": {
        "input_tokens": 24735,
        "cached_input_tokens": 24576,
        "output_tokens": 314,
        "reasoning_output_tokens": 262,
        "total_tokens": 25049
      }
    }
```

## Notes
Any additional context or explanations for checkbox items that don't fully apply


foxie-huang:
This is a well designed task that I'd lean to approve. But some suggestions for improvements before merging: 
**Realism:**

1. **Use different securities for at least 2-3 sectors in the portfolio vs benchmark.** This is the big one. Right now selection and interaction are identically zero, which means 2/3 of the Brinson-Fachler framework is untested. If the portfolio held, say, VGT instead of XLK for tech, and IYF instead of XLF for financials, you'd get nonzero selection and interaction effects. The data file would just need a few extra columns of daily prices. This would turn the task from "BF allocation only" into an actual full BF attribution.

2. **Specify that prices are adjusted closes.** The instruction is silent on dividend treatment. Real sector ETFs pay dividends — XLE's yield is ~3.5%. If an agent uses unadjusted prices, returns will be wrong by the dividend amount. Either state "prices are dividend-adjusted closes" so it's clear, or provide raw prices + dividend data and make the agent handle it. The former is simple realism; the latter is a difficulty upgrade.

3. **Let benchmark weights drift too.** Real S&P 500 sector weights change monthly as market caps move. You could provide benchmark weights at the start of each quarter (matching the rebalance dates), or even monthly. This eliminates the unrealistic "fixed benchmark" assumption and adds a subtle implementation detail — the agent must use the right benchmark weights for each month.

4. **Replace `risk_free_annual / 12` with actual monthly T-bill rates.** Provide a small `cash_rates.csv` with 12 monthly rates (or even just 4 quarterly). This is trivial extra data but eliminates the simple-rate-vs-compound-rate ambiguity and reflects how real cash returns are computed.

**Distinguishing ability:**

1. **Test post-rebalance months (Apr, Jul, Oct).** These are the months where weights snap back to target. An agent with a wrong drift implementation might produce correct results at rebalance months (because weights reset) but diverge in between — or vice versa, get the reset wrong. Currently tested months (Jan, Feb, Jun, Sep, Dec) include only one post-rebalance month (Jan). Adding Apr and Oct would directly probe the rebalance logic.

2. **Add a weight snapshot test.** Test the beginning-of-month weights for a specific mid-quarter month (e.g., "XLK weight at the start of March should be ~0.358"). This directly verifies drift computation independently from the BF formula. Right now drift errors are only caught indirectly through their effect on allocation. A direct weight test would localize failures.

3. **Test one month's per-sector allocation breakdown, not just annual totals.** The per-sector tests (`test_sector_allocation`) sum 12 months. An agent could get the annual sum right through compensating errors across months. Testing, say, March's XLK allocation specifically would catch agents that got the formula right but the drift wrong, or vice versa.

4. **Add a sign/sanity test for CASH allocation.** Cash at 5.25% annual in a year where the benchmark returned ~19.7% means cash is a massive drag. Test that `sector_allocation["CASH"] < -0.01` — this catches agents that get the sign wrong on the BF formula for cash (a common mistake since cash has zero benchmark weight, so the formula simplifies differently).

5. **Drop the two trivially-zero tests if you don't add different securities.** `test_selection_is_zero` and `test_interaction_is_zero` contribute no discriminative signal — they only catch agents that hallucinate nonzero values. If you adopt suggestion #1 from the realism section (different securities), these tests become meaningful. If not, they're padding the test count without adding information.

boqiny:
@foxie-huang 

Thanks for the review! I addressed everything: the portfolio now uses VGT, IYF, and VHT instead of XLK, XLF, XLV for three sectors, so selection and interaction are no longer trivially zero. Benchmark weights drift quarterly using actual 2024 market movements, and the fixed risk free rate is replaced with monthly T-bill rates from a new cash_rates.csv. Prices are labeled as dividend adjusted closes. 

For the testing side, I added post-rebalance month tests for Apr, Jul, and Oct, a weight snapshot at the start of March to verify drift directly, a per-sector monthly breakdown test for XLK in March, and a cash sign check. The old zero-value selection and interaction tests are replaced with real ones across the three differing sectors. All new price data and rates are pulled from Yahoo Finance. Also the corresponding results has been updated in the PR description.

PR #128, opened 2026-04-10
feat: add etf-cross-asset-lead-lag task (medium, cross-asset-analysis)

xinlan-technology:
- [x] I ran `harbor tasks check tasks/etf-cross-asset-lead-lag` on my new task and ensured all checks pass
  - [x] `instruction.md` does NOT mention which skills to use (agents must discover skills themselves)
  - [x] It is hard for the agent to cheat (e.g., editing data files, looking inside files for solution strings, etc.)
  - [x] My `solution/solve.sh` was written by a human (with minimal help from a language model)
  - [x] If external dependencies are used, versions are pinned for reproducibility
  - [x] If the agent produces structured data (API, JSON, CSV, etc.), the exact schema is documented in instruction.md or a referenced spec file
  - [x] Model Validation: I have verified this task using at least one model from the Frontier Tier (e.g., Claude Opus 4.6, GPT-5.4, or Gemini 3 Pro) in the Agent Runs
  section below
  - [x] I have provided a root cause analysis for every unit test failure for each agent run in the Agent Runs section below

  ---

  **Inspired by:** ML4T Case Study #1 — ETF Cross-Asset Exposures (https://ml3t.applikuapp.com/case-studies/)

  ## Summary

  Analyze cross-asset lead-lag relationships among 10 ETFs (SPY, QQQ, IWM, EFA, TLT, IEF, LQD, GLD, USO, VNQ) using lagged correlation asymmetry. The pipeline splits returns into in-sample (2018-2021) and out-of-sample (2022-2023), identifies the strongest lead-lag pairs, validates their persistence out-of-sample, removes the SPY market factor via OLS to find idiosyncratic lead-lag signals, and checks residual pair stability across 2022 and 2023 subperiods. Uses real Yahoo Finance daily ETF price data.

  ## Task Metadata
  | Field | Value |
  |-------|-------|
  | **Task ID** | `etf-cross-asset-lead-lag` |
  | **Difficulty** | Medium |
  | **Category** | cross-asset-analysis |
  | **Data Source of Input Files** | Yahoo Finance (yfinance), 10 ETF adjusted close prices 2018-2023 |

  ## Agent Performance
  | Oracle | — | 1.000 | 20/20 | — |
  | Finance-Zero | gemini-2.5-flash | 0.000 | 0/20 | SyntaxError (markdown fence in output) |
  | claude-code | claude-opus-4-6 | 0.000 | 18/20 | Used simple returns instead of log returns |
  | codex | gpt-5.4 | 0.000 | 15/20 | CSV column format error, CSV not sorted, solution.json rounding |

  ## Screenshots / Evidence

  ### Oracle Run
oracle

  ### Agent Runs

  #### Claude Opus 4.6 — 18/20 passed
opus-4-6

  **Root cause:** Used simple returns `(P_t - P_{t-1}) / P_{t-1}` instead of log returns `ln(P_t / P_{t-1})`. This produced `spy_annualized_return = 0.1598` (expected ~0.1121 for log returns). The instruction explicitly requires log returns, but Opus defaulted to simple returns. All other computations (lead-lag asymmetry, market model, residual analysis, subperiod validation) were correct — only the return computation was wrong.

  #### Codex GPT-5.4 — 15/20 passed
gpt-5 4

  **Root cause 1:** CSV column naming — used `lag_0` instead of `lag_+0` for the zero-lag column. The oracle uses Python's `f"lag_{k:+d}"` format which produces `lag_+0` for k=0. GPT omitted the explicit positive sign.

  **Root cause 2:** Both `lead_lag_results.csv` and `residual_lead_lag_results.csv` were not sorted by `abs(asymmetry)` descending, despite the instruction requiring sorted output.

  **Root cause 3:** `solution.json` intermediate `tlt_annualized_return` was rounded to 6 decimal places (`-0.016839`) instead of preserving full precision (`-0.016839366...`). The instruction states "do not round it again separately."

  #### Finance-Zero (gemini-2.5-flash) — 0/20 passed
finance-zero

  **Root cause:** Generated markdown code fences in output, causing immediate SyntaxError. No output files produced.

  ## Notes
  This task does not include skills. All behavior is tested via pytest assertions in `test_outputs.py` (20 test functions). Data sourced from Yahoo Finance via `yfinance` library — no API key required.


foxie-huang:
The test results look good and the task is realistic. I'd approve pending the following suggested improvements:

**Realism**

**1. Add a significance gate before ranking.**

The pipeline ranks all 45 pairs by raw asymmetry magnitude, but with ~1000 in-sample observations many small asymmetries are noise. The task's own data confirms this — 0% of top raw pairs persist OOS. Require the agent to compute a standard error for the asymmetry via Fisher z-transform and only include pairs where the asymmetry is significant at 5%. If fewer than num_top_pairs pass the filter, report however many do. This transforms the task from "sort a noisy list" to "identify statistically meaningful signals."

**2. Use peak-lag asymmetry instead of lag-1 only.**

Define asymmetry as the difference between the maximum positive-lag correlation and the maximum negative-lag correlation across the full ±5 window. The current lag-1-only definition misses lead-lag relationships that operate at longer horizons and wastes the full correlogram the task already computes. Practitioners look at the full shape, not a single point.

**3. Make the agent interpret the OOS result.**

Add a signal_assessment output with two booleans: whether the raw signal persists OOS (sign consistency > 50%) and whether residualization improves persistence. These are trivial comparisons but they force the agent to connect pipeline outputs to a conclusion. Right now the task computes numbers and stops — a real researcher would draw an inference.

**4. Compute annualized return correctly.**

Replace mean × 252 with the sum of log returns divided by the number of years. For log returns over a multi-year sample, the sum is the exact log compound return. The mean×252 approach conflates "expected annual return" with "realized annualized return" — a known source of confusion that the task should avoid rather than perpetuate.

**Distinguishing Ability**

**1. Relax intermediate tolerances from 1e-12 to 1e-6. (Highest priority)**

The current intermediates test checks values to 12 decimal places. This tests floating-point determinism across platforms, not finance. Different numpy BLAS backends or operation ordering can produce values that differ at the 10th decimal place while being financially identical. The log-vs-simple-returns discriminator still works perfectly at 1e-6 — the gap is ~0.05.

**2. Test rank bands instead of exact ranks for subperiod validation.**

If two pairs have asymmetries differing by 1e-7, different sort implementations could swap their ranks. Replace exact rank assertions with band tests — "still top-ranked" (≤3) instead of "exactly rank 1", "dropped significantly" (≥20) instead of "exactly rank 27". This preserves the signal while being robust to near-ties.

**3. Remove the lag_+0 column formatting trap.**

GPT-5.4 failed because it wrote lag_0 instead of lag_+0 for the zero-lag column. This is Python string formatting trivia, not quantitative finance. Either accept both forms or change the spec to use lag_0. The column content is already tested elsewhere via corr_lag0.

**4. Add a test that residual persistence exceeds raw persistence.**

The core hypothesis of the entire pipeline is that removing market exposure should improve OOS signal quality. Test that the residual same-sign count is at least as high as the raw same-sign count. This is a property test — robust to data changes and directly checks financial reasoning rather than numerical output.

**5. Add a test that residualization changes the top pair.**

If removing SPY exposure doesn't change which pair ranks first, the residualization isn't doing anything meaningful. Test that the top raw pair and top residual pair differ. This is another property test that checks understanding of why we residualize, not just whether the formulas were implemented correctly.

**6. Replace one consistency test with a cross-pipeline property test.**

The current consistency tests (solution.json matches results.json) just verify the same value appears in two places — they test bookkeeping, not finance. Replace one with a test that high-beta ETFs show larger changes in lead-lag asymmetry after residualization than low-beta ETFs, since more of their raw correlation structure was driven by shared market exposure. This adds genuine financial reasoning to the test suite.

xinlan-technology:
> The test results look good and the task is realistic. I'd approve pending the following suggested improvements:
> 
> **Realism**
> 
> **1. Add a significance gate before ranking.**
> 
> The pipeline ranks all 45 pairs by raw asymmetry magnitude, but with ~1000 in-sample observations many small asymmetries are noise. The task's own data confirms this — 0% of top raw pairs persist OOS. Require the agent to compute a standard error for the asymmetry via Fisher z-transform and only include pairs where the asymmetry is significant at 5%. If fewer than num_top_pairs pass the filter, report however many do. This transforms the task from "sort a noisy list" to "identify statistically meaningful signals."
> 
> **2. Use peak-lag asymmetry instead of lag-1 only.**
> 
> Define asymmetry as the difference between the maximum positive-lag correlation and the maximum negative-lag correlation across the full ±5 window. The current lag-1-only definition misses lead-lag relationships that operate at longer horizons and wastes the full correlogram the task already computes. Practitioners look at the full shape, not a single point.
> 
> **3. Make the agent interpret the OOS result.**
> 
> Add a signal_assessment output with two booleans: whether the raw signal persists OOS (sign consistency > 50%) and whether residualization improves persistence. These are trivial comparisons but they force the agent to connect pipeline outputs to a conclusion. Right now the task computes numbers and stops — a real researcher would draw an inference.
> 
> **4. Compute annualized return correctly.**
> 
> Replace mean × 252 with the sum of log returns divided by the number of years. For log returns over a multi-year sample, the sum is the exact log compound return. The mean×252 approach conflates "expected annual return" with "realized annualized return" — a known source of confusion that the task should avoid rather than perpetuate.
> 
> **Distinguishing Ability**
> 
> **1. Relax intermediate tolerances from 1e-12 to 1e-6. (Highest priority)**
> 
> The current intermediates test checks values to 12 decimal places. This tests floating-point determinism across platforms, not finance. Different numpy BLAS backends or operation ordering can produce values that differ at the 10th decimal place while being financially identical. The log-vs-simple-returns discriminator still works perfectly at 1e-6 — the gap is ~0.05.
> 
> **2. Test rank bands instead of exact ranks for subperiod validation.**
> 
> If two pairs have asymmetries differing by 1e-7, different sort implementations could swap their ranks. Replace exact rank assertions with band tests — "still top-ranked" (≤3) instead of "exactly rank 1", "dropped significantly" (≥20) instead of "exactly rank 27". This preserves the signal while being robust to near-ties.
> 
> **3. Remove the lag_+0 column formatting trap.**
> 
> GPT-5.4 failed because it wrote lag_0 instead of lag_+0 for the zero-lag column. This is Python string formatting trivia, not quantitative finance. Either accept both forms or change the spec to use lag_0. The column content is already tested elsewhere via corr_lag0.
> 
> **4. Add a test that residual persistence exceeds raw persistence.**
> 
> The core hypothesis of the entire pipeline is that removing market exposure should improve OOS signal quality. Test that the residual same-sign count is at least as high as the raw same-sign count. This is a property test — robust to data changes and directly checks financial reasoning rather than numerical output.
> 
> **5. Add a test that residualization changes the top pair.**
> 
> If removing SPY exposure doesn't change which pair ranks first, the residualization isn't doing anything meaningful. Test that the top raw pair and top residual pair differ. This is another property test that checks understanding of why we residualize, not just whether the formulas were implemented correctly.
> 
> **6. Replace one consistency test with a cross-pipeline property test.**
> 
> The current consistency tests (solution.json matches results.json) just verify the same value appears in two places — they test bookkeeping, not finance. Replace one with a test that high-beta ETFs show larger changes in lead-lag asymmetry after residualization than low-beta ETFs, since more of their raw correlation structure was driven by shared market exposure. This adds genuine financial reasoning to the test suite.



> The test results look good and the task is realistic. I'd approve pending the following suggested improvements:
> 
> **Realism**
> 
> **1. Add a significance gate before ranking.**
> 
> The pipeline ranks all 45 pairs by raw asymmetry magnitude, but with ~1000 in-sample observations many small asymmetries are noise. The task's own data confirms this — 0% of top raw pairs persist OOS. Require the agent to compute a standard error for the asymmetry via Fisher z-transform and only include pairs where the asymmetry is significant at 5%. If fewer than num_top_pairs pass the filter, report however many do. This transforms the task from "sort a noisy list" to "identify statistically meaningful signals."
> 
> **2. Use peak-lag asymmetry instead of lag-1 only.**
> 
> Define asymmetry as the difference between the maximum positive-lag correlation and the maximum negative-lag correlation across the full ±5 window. The current lag-1-only definition misses lead-lag relationships that operate at longer horizons and wastes the full correlogram the task already computes. Practitioners look at the full shape, not a single point.
> 
> **3. Make the agent interpret the OOS result.**
> 
> Add a signal_assessment output with two booleans: whether the raw signal persists OOS (sign consistency > 50%) and whether residualization improves persistence. These are trivial comparisons but they force the agent to connect pipeline outputs to a conclusion. Right now the task computes numbers and stops — a real researcher would draw an inference.
> 
> **4. Compute annualized return correctly.**
> 
> Replace mean × 252 with the sum of log returns divided by the number of years. For log returns over a multi-year sample, the sum is the exact log compound return. The mean×252 approach conflates "expected annual return" with "realized annualized return" — a known source of confusion that the task should avoid rather than perpetuate.
> 
> **Distinguishing Ability**
> 
> **1. Relax intermediate tolerances from 1e-12 to 1e-6. (Highest priority)**
> 
> The current intermediates test checks values to 12 decimal places. This tests floating-point determinism across platforms, not finance. Different numpy BLAS backends or operation ordering can produce values that differ at the 10th decimal place while being financially identical. The log-vs-simple-returns discriminator still works perfectly at 1e-6 — the gap is ~0.05.
> 
> **2. Test rank bands instead of exact ranks for subperiod validation.**
> 
> If two pairs have asymmetries differing by 1e-7, different sort implementations could swap their ranks. Replace exact rank assertions with band tests — "still top-ranked" (≤3) instead of "exactly rank 1", "dropped significantly" (≥20) instead of "exactly rank 27". This preserves the signal while being robust to near-ties.
> 
> **3. Remove the lag_+0 column formatting trap.**
> 
> GPT-5.4 failed because it wrote lag_0 instead of lag_+0 for the zero-lag column. This is Python string formatting trivia, not quantitative finance. Either accept both forms or change the spec to use lag_0. The column content is already tested elsewhere via corr_lag0.
> 
> **4. Add a test that residual persistence exceeds raw persistence.**
> 
> The core hypothesis of the entire pipeline is that removing market exposure should improve OOS signal quality. Test that the residual same-sign count is at least as high as the raw same-sign count. This is a property test — robust to data changes and directly checks financial reasoning rather than numerical output.
> 
> **5. Add a test that residualization changes the top pair.**
> 
> If removing SPY exposure doesn't change which pair ranks first, the residualization isn't doing anything meaningful. Test that the top raw pair and top residual pair differ. This is another property test that checks understanding of why we residualize, not just whether the formulas were implemented correctly.
> 
> **6. Replace one consistency test with a cross-pipeline property test.**
> 
> The current consistency tests (solution.json matches results.json) just verify the same value appears in two places — they test bookkeeping, not finance. Replace one with a test that high-beta ETFs show larger changes in lead-lag asymmetry after residualization than low-beta ETFs, since more of their raw correlation structure was driven by shared market exposure. This adds genuine financial reasoning to the test suite.

@foxie-huang 

Thanks for the thorough review. All changes are in commit `93ff709`.

Implemented as suggested:
- D1: Relaxed all tolerances from `1e-12` to `1e-6`
- D2: Replaced exact rank assertions with rank-band tests (`≤3` / `≤10` / `≥20`)
- D3: Changed `lag_+0` to `lag_0` and updated the spec, CSVs, and tests
- D5: Added a property test that the top raw pair differs from the top residual pair
- R1: Added a Fisher-z significance gate (5% two-sided, `1.96 * sqrt(2/(n-3))`)
- R2: Changed asymmetry to peak-lag asymmetry: `max(corr[+1..+5]) - max(corr[-1..-5])`
- R3: Added `signal_assessment` with `raw_persists_oos` and `residual_improves_persistence`
- R4: Changed annualized return to calendar-year annualization: `sum(log_returns) / ((last_date - first_date).days / 365.25)`

Implemented differently from the literal suggestion:
- D4 (`residual persistence ≥ raw persistence`): I did not add this as a hard assertion. After applying R1+R2, this dataset gives raw persistence = `3/5` (60%) and residual persistence = `2/4` (50%), so the hypothesis does not hold empirically. Instead of hardcoding a false property, `signal_assessment.residual_improves_persistence` correctly reports `False`, and the tests verify that the agent computes this boolean correctly. I’m happy to redesign the task to make this hypothesis hold, but that would be a larger-scope change involving the sample window or pair-selection setup.
- D6 (high-beta vs low-beta asymmetry change): I implemented this at the pair level using `|beta_a * beta_b|`, since the market-driven covariance term for a pair is proportional to `beta_a * beta_b * Var(SPY)`. I also kept the existing consistency test rather than replacing it, because it caught a real agent failure in the original PR.

Agent results (24 tests):
- Oracle: `24/24`
- Claude-code (opus-4-6): `23/24`
- Codex (gpt-5.4): `24/24` on this run
- Finance-Zero (gemini-2.5-flash): `0/24`

The remaining frontier-agent failure mode is intermittent over-rounding of `avg_abs_asymmetry_ratio` (about `3.2e-6` vs a `1e-6` tolerance), which matches the instruction-level discriminator rather than a formatting trap.

yunfei67 (approved):
Thanks for the work here! The task is well-scoped for medium difficulty — the pipeline covers a realistic research workflow (in-sample discovery → OOS validation → market-factor residualization → subperiod stability). The significance gate, peak-lag asymmetry, and design-validity tests (D5/D6) are particularly nice additions. LGTM.

PR #132, opened 2026-04-15
Add task dirty-gap-momentum-aapl (easy)

Jiahao-Xie-86:
Made-with: Cursor

---

## Summary
This PR adds `dirty-gap-momentum-aapl`, an easy backtesting task built around dirty real AAPL adjusted-price data. The agent must clean duplicated and partially missing rows, reconstruct `adj_open` where possible, run an overnight gap-momentum backtest with explicit trading rules and transaction costs, and emit exact JSON/CSV/HTML artifacts.

The task tests basic but realistic quant workflow skills: data cleaning, adjusted-price handling, signal construction from lagged prices, intraday trade accounting, and exact structured-output compliance.

## Task Metadata
| Field | Value |
|-------|-------|
| **Task ID** | `dirty-gap-momentum-aapl` |
| **Difficulty** | **Easy** — single-asset strategy, simple signal logic, no optimization or factor model, but still requires deterministic cleaning and correct adjusted-price accounting |
| **Category** | `backtesting` |
| **Data Source of Input Files** | Yahoo Finance-style market data |

## Agent Performance
| Run | Model | Tests Passed | Reward |
|---|---|---:|---:|
| Oracle | -- | 19 / 19 | 1.0 |
| Finance-Zero | `gemini/gemini-2.5-flash` | 0 / 19 | 0.0 |
| Gemini CLI | `google/gemini-3.1-pro-preview` | 19 / 19 | 1.0 |

### Harbor tasks check
dirty-gap-momentum-aapl-harbor-check

### Oracle Run
dirty-gap-momentum-aapl-harbor-oracle

### Agent Runs
####Finance-Zero (`gemini/gemini-2.5-flash`)
dirty-gap-momentum-aapl-harbor-finance-zero-gemini-2 5-flash
- Failure summary: `0 / 19` tests passed, `reward = 0.0`
- Root cause analysis:
  - Finance-Zero generated invalid Python with an unterminated string literal.
  - As a result, the script never created the required output artifacts.
  - The verifier then failed all 19 tests because `results.json`, `cleaned_prices.csv`, `trades.csv`, `daily_portfolio.csv`, and `cumulative_returns.html` were missing.
- `final_metrics` from `agent/trajectory.json`: not available for Finance-Zero in this run scaffold.

#### Gemini CLI (`google/gemini-3.1-pro-preview`)
dirty-gap-momentum-aapl-harbor-gemini-cli-google-gemini-3 1-pro-preview
- Verifier output: `19 / 19` tests passed, `reward = 1.0`
- Failure analysis: none; the agent passed the full verifier.
- `final_metrics` from `agent/trajectory.json`:
  - `total_prompt_tokens`: `399980`
  - `total_completion_tokens`: `13864`
  - `total_cached_tokens`: `336555`
  - `total_steps`: `18`

## Notes
- This task does not ship any `environment/skills`, so the skills-specific checklist items are intentionally marked not applicable.


yunfei67 (commented):
1. Finance-Zero calibration result is not informative.
Finance-Zero failed 0/19 because it generated a markdown code fence that caused a SyntaxError, not because the task was too hard. Per the review guideline, Finance-Zero must fail due to task difficulty — this run doesn't confirm that. Please re-run Finance-Zero with a more stable model (e.g. gemini/gemini-2.0-flash) to get a valid difficulty calibration signal.

2. profit_factor edge case is undocumented.
instruction.md does not specify what to report when gross_loss = 0 (no losing trades). The oracle returns None in that case, but test_results_numerics uses np.isclose, which raises a TypeError on None rather than producing a meaningful test failure. The current dataset doesn't trigger this, but please document the expected behavior (e.g. null, omit the field, or return Infinity) so agents have a clear contract to follow.

foxie-huang:
I think this task is more about data handling rather than finance. The following improvements are suggested and we'll see how the task runs after the changes are made. 

**Improving realism**
**1. Let the agent derive at least one convention instead of transcribing it.**
Replace the spelled-out imputation formula adj_open = open * adj_close / close with a situational description ("reconstruct adj_open using the adjustment ratio implied by the same row"). Same oracle answer, but now the task distinguishes agents that understand adjusted prices from agents that copy formulas. Real analysts derive; specs in the real world are incomplete.

**2. Add one frictional detail that backtests actually have.**
Pick one: (a) a borrow fee on the short leg (e.g., 25 bps annualized applied on intraday notional), (b) slippage as a half-spread on entry and exit, or (c) a volume sanity check that caps shares at some fraction of that day's volume. Any one of these pushes the task from "toy textbook" to "resembles something a junior quant would actually be asked to code" without multiplying the spec surface.

**3. Include at least one dirty-data pathology that requires judgment.**
Right now the dirty data is mechanically fixable (dupes → keep last; missing adj_open → one formula). Add one ambiguous row — e.g., a split-day where adj_close jumps but adj_open doesn't, or a zero-volume day with stale prices — and let the instruction describe the symptom ("rows where OHLC is inconsistent with an adjustment event"). The oracle still knows the answer; the agent has to recognize the pathology.

**Improving discrimination**
**1. Re-calibrate with a three-point gradient.**
Add a stable weak baseline (GPT-4.1-mini or Haiku 4.5 — not flash-with-codegen-bugs) and a mid-tier model (Sonnet 4.5 or Gemini 2.0 Flash). A defensible task should show roughly monotone scores like 3 / 11 / 19. Without that middle point, the PR has shown solvability, not discrimination.

**2. Add derived-property tests alongside exact-numeric tests.**
Keep the 13 parametrized oracle-value tests, but add 3–4 relationship tests that a capable model can pass even if it picked a different convention: total_return ≈ final_capital/100_000 − 1, long_trades + short_trades == num_trades, cumulative_return[-1] ≈ 1 + total_return, max_drawdown ≥ 0, abs(portfolio_value.diff().sum() − (final − initial)) < 1e-6. This creates a partial-credit floor for models that get the logic right but disagree on ddof or annualization.

**3. Soften the exact-count assertions into bands, and split "structure" from "numerics" cleanly.**
assert len(trades) == 305 currently conflates three independent spec details (dedupe rule, threshold inclusivity, shares==0 skip). Replace with a band (300 ≤ n ≤ 310) plus separate tests that each isolate one rule — e.g., a test that checks "no trade on days where gap_return is exactly 0.009" (threshold strictness), "no duplicate dates in trades" (dedupe), "no trade with shares == 0" (sizing rule). Isolated checks give diagnostic signal; the aggregate count gives none.

Jiahao-Xie-86:
Thanks. I updated `dirty-gap-momentum-aapl` to address the reviewer feedback and reran the round-2 evaluation suite for the revised task.

**What changed**

I made four substantive changes:

1. **Adjusted-price reconstruction is now less formula-copyable**
- In `tasks/dirty-gap-momentum-aapl/instruction.md`, I replaced the explicit `adj_open = open * adj_close / close` formula with a situational requirement: reconstruct `adj_open` using the adjustment ratio implied by the same row.

2. **Added a realistic friction**
- I added a short borrow fee of `25 bps` annualized on short entry notional.
- This is documented in `tasks/dirty-gap-momentum-aapl/instruction.md` and implemented in `tasks/dirty-gap-momentum-aapl/solution/solve.sh`.

3. **Added a judgment-requiring dirty-data pathology**
- I inserted a stale zero-volume row into `tasks/dirty-gap-momentum-aapl/environment/data/aapl_prices_dirty.csv`.
- The instruction now explicitly requires dropping rows with zero volume and unchanged unadjusted/adjusted prices relative to the previous retained row.
- The verifier directly checks that this stale row is removed.

4. **Improved verifier discrimination and contract clarity**
- I documented the `profit_factor` edge case in `tasks/dirty-gap-momentum-aapl/instruction.md`: if `gross_loss == 0`, report `profit_factor` as `null`.
- I updated `tasks/dirty-gap-momentum-aapl/tests/test_outputs.py` so that edge case is handled safely.
- I also made the verifier more diagnostic by:
  - adding direct stale-row removal coverage,
  - replacing the brittle exact aggregate trade-count assertion with a band,
  - adding a threshold-isolation test for flat days,
  - adding derived consistency checks such as:
    - `long_trades + short_trades == num_trades`
    - `total_return == final_capital / 100000 - 1`
    - final cumulative return matches final capital
    - sum of trade PnL matches final minus initial capital
    
    **Updated results**

Oracle:
dirty-gap-momentum-aapl-harbor-oracle
- Reward: `1.0`
- Tests passed: `11 / 11`

Gemini CLI (`google/gemini-2.5-flash`):
dirty-gap-momentum-aapl-harbor-gemini-cli-google-gemini-2 5-flash
- Reward: `0.0`
- Tests passed: `6 / 11`
- Root cause summary:
  - the model produced a partially correct solution but over-dropped rows during cleaning (`1022` clean rows instead of `1026`),
  - that caused the direct cleaning test to fail,
  - and the downstream portfolio / numeric checks drifted accordingly,
  - it also missed the expected Plotly output details checked by the verifier.

Gemini CLI (`google/gemini-3.1-pro-preview`):
dirty-gap-momentum-aapl-harbor-gemini-cli-google-gemini-3 1-pro-preview
- Reward: `1.0`
- Tests passed: `11 / 11`

PR #135, opened 2026-04-15
Add task: prediction-markets-cross-venue-dislocation

xushenbo:
# PR Description

- [ ]  I ran `harbor tasks check tasks/<task-id>` on my new task and ensured all checks pass
- [ ]  My `instruction.md` was written by a human
- [ ]  My test cases have informative docstrings that describe which behavior they check
- [ ]  My `task.toml` was written by a human
- [ ]  My `solution/solve.sh` was written by a human (with minimal help from a language model)
- [ ]  If external dependencies are used, versions are pinned for reproducibility

---

## Summary
Adds a new hard task: `prediction-markets-cross-venue-dislocation`.

This task uses a frozen derived prediction-market dataset built from matched Kalshi and Polymarket 2024 Republican Senate seat markets. The agent must clean and synchronize cross-venue snapshots, apply fee-aware locked-edge logic, enforce trading and capital constraints, and produce exact outputs for portfolio metrics, trade logs, daily equity, and intermediate checkpoints.

## Task Metadata
| Field | Value |
|-------|-------|
| **Task ID** | `prediction-markets-cross-venue-dislocation` |
| **Difficulty** | Hard. Oracle passes, while two strong frontier runs failed for task reasons. |
| **Category** | `cross-domain` |
| **Data Source of Input Files** | Public Kalshi historical market/candlestick endpoints and public Polymarket Gamma/CLOB market-data endpoints, frozen into a matched Republican Senate seats 2024 package. |

## Agent Performance
- Oracle run passes with reward `1.0`.
- `codex` with `openai/gpt-5.4` fails with reward `0.0`.
- `codex` with `openai/o3` fails with reward `0.0`.

## Screenshots / Evidence
codex-gpt-5 4-fail
codex-o3-fail
oracle-pass

### Harbor tasks check
Not included. In the local Harbor install used for this task, `harbor check` still requested `ANTHROPIC_API_KEY` even when an OpenAI model was specified.

### Oracle Run
Oracle passes with reward `1.0`.

### Agent Runs
| Agent | Model | Result | Reward |
|---|---|---|---|
| `oracle` | n/a | pass | `1.0` |
| `codex` | `openai/gpt-5.4` | fail | `0.0` |
| `codex` | `openai/o3` | fail | `0.0` |

Failure analysis:
- `codex / openai/gpt-5.4`: the agent produced the required output files, but failed the verifier on quantitative/spec mismatches. The first failing assertion was on `results["net_edge_bps_mean"]`, where the produced value was `205.215` instead of the expected `205.21375`. The run also diverged from the benchmark contract on candidate-signal counting and annualized performance calculations.
- `codex / openai/o3`: the agent failed to produce the required `/app/output` artifacts (`results.json`, `solution.json`, `trade_log.csv`, and `daily_equity.csv`). Verification therefore failed immediately on missing files.

## Notes
The benchmark inputs are a frozen derived dataset created from public Kalshi historical market/candlestick data and public Polymarket market-data endpoints. The final committed files are derived benchmark inputs for a matched set of exact 2024 Republican Senate seat-count markets across the two venues.

The task does not fetch live data at runtime.


PQCat:
Claude-haiku-4-5, Round: 1, Turns: 24/40
Time: 337.5s, Tokens: 84263in/261out
Test: 3 passed in 22.23s
Result: PASS

Dongzhikang (commented):
## Review: Add task: prediction-markets-cross-venue-dislocation
Reviewer: Grim 🔍 | Date: 2026-04-18

### Findings

**[MAJOR] The verifier exposes too much of the reference run**  
`tasks/prediction-markets-cross-venue-dislocation/tests/test_outputs.py`

The current tests hardcode a large amount of real-data output structure, including exact summary values, intermediate values, selected pair ordering, trade-log details, and daily-equity checkpoints.

That makes the task more snapshot-matching than spec-validation, and it weakens benchmark integrity if an agent can read `/tests`.

A more robust approach would be:
- keep a small number of coarse final anchors,
- validate schema + accounting identities + trading constraints directly,
- and avoid exposing so much of the full reference path.

**[MINOR] The early-close trading window should be stated more explicitly**  
`tasks/prediction-markets-cross-venue-dislocation/instruction.md`

Because the task stops signal generation at the earlier venue close, the effective trading window is materially shorter than the full resolution horizon. If that is intentional, I would make it explicit in the spec that once either venue is no longer tradable, the pair is no longer eligible for new entries even if the other venue continues trading.

### Verdict
REQUEST CHANGES

The task idea is good, but the verifier currently reveals too much of the reference run and should be tightened before merge.

xushenbo:
Thanks. I addressed both review items.

For the verifier leakage concern, I tightened the real-data test so it no longer exposes a large portion of the reference path. The public verifier now keeps only a small number of coarse anchors and validates schema, accounting identities, trading constraints, synchronized-snapshot consistency, and cross-file consistency derived from the task inputs. Exact path-style checks remain on the small synthetic fixtures, where they are useful for edge-case coverage, but the real-data test is no longer snapshot-matching the full reference run.

For the spec wording, I updated the instruction to make the entry window explicit: a pair is eligible for new entries only while `snapshot_ts_utc < min(kalshi_close_ts_utc, polymarket_close_ts_utc, resolution_ts_utc)`. I also clarified that once either venue has closed, the pair is no longer eligible for new entries even if the other venue remains open, while already-open trades are still held to `resolution_ts_utc`.

I also cleaned up the verifier for reproducibility and coverage while I was there:

- pinned the task-specific verifier dependencies
- pinned the task base image by digest
- moved verifier deps into the task image instead of downloading them at test time
- added explicit coverage for the invalid/missing price cleaning path

Validation on the promoted `tasks/` copy now passes:

- `harbor check tasks/prediction-markets-cross-venue-dislocation`
- `harbor run --path tasks/prediction-markets-cross-venue-dislocation --agent oracle` -> reward `1.0`
- `harbor run --path tasks/prediction-markets-cross-venue-dislocation --agent codex --model openai/o3` -> `Errors: 0`, reward `0.0`
- `harbor run --path tasks/prediction-markets-cross-venue-dislocation --agent codex --model openai/gpt-5.4` -> `Errors: 0`, reward `0.0`

So the task now clears the review concerns and still appears nontrivial for multiple frontier models.
codex-gpt-5 4-fail
codex-o3-fail
oracle-pass

beckybyte:
Your docker file has problem and `solution/solve.sh` has grammar mistake. Please correct and be careful.

beckybyte:
hi, I helped you debug a little bit, and here's some results 

```
Both the `run_oracle()` and `run_oracle_from_data_dir()` functions execute `bash SOLVE_SH`—that is, they run the Oracle's `solve.sh` script. All three `test_*` functions call upon these two functions.

Crucially, none of the existing tests verify the output generated by the agent within the `/app/output/` directory.

Consequently, once the agent finishes running, the testing framework immediately attempts to locate `solve.sh`; failing to find it, the tests fail outright, and any output files generated by the agent are completely ignored.

The author needs to add tests similar to the following:

```python
def test_output_files_exist():
assert (Path("/app/output") / "results.json").exists()
assert (Path("/app/output") / "trade_log.csv").exists()

def test_results_schema():
results = json.load(open("/app/output/results.json"))
assert "num_trades" in results
assert results["num_trades"] >= 0
...
```

Current Status: The Oracle runs successfully, but the agent consistently yields a result of 0. This PR cannot be merged yet.

Due to our time is limited, I'll test your PR at most twice. Thanks for you understanding.

xushenbo:
Thank you so much for your help debugging! I updated my code and pushed again. Could you please help check with /test-deploy ? Thanks!

beckybyte:
Also I fixed docker bug before. If possible, find another device and check your results.

xushenbo:
Fingers crossed this works!

beckybyte:
> Also I fixed docker bug before. If possible, find another device and check your results.

You need to fix that.

Here's the notice from Claude before your last commit.

```
remind the author that even if the checkout succeeds, the oracle will still fail because the SOLVE_SH path in test_outputs.py is incorrect (/app/solution/solve.sh instead of /solution/solve.sh).
They need to be told to:
1. Change
SOLVE_SH = TASK_DIR / "solution" / "solve.sh"
to
SOLVE_SH = Path("/solution/solve.sh")
2. Add agent tests (checking files under /app/output/), because right now all tests are oracle tests, so the agent can never pass.

```

xushenbo:
Thank you for the update!

The verifier no longer reruns solve.sh in deploy mode. It now reads agent produced files from /app/output, adds explicit output-file existence and schema checks, and keeps the full invariant-based validation on those outputs. I also fixed the deploy-time real-data path resolution so the verifier reads from /app/data instead of the wrong /app/environment/data.

foxie-huang:
Thanks for this PR — clear `instruction.md`, well-tested oracle, real
frozen Kalshi + Polymarket data, and a meaningful coverage addition
(binary-event prediction markets and cross-venue mechanics aren't
currently in the catalog). Happy to merge after a few changes — the
main one is in the verifier scoring design, plus one realism point
on execution.

### Required changes (blocking)

**1. Heterogeneous tolerances + stage-grouped partial credit.**

`test_outputs.py` uses `abs=1e-6` on essentially every numerical
assertion. Combined with the deterministic 14-step pipeline, this
produces a near-binary outcome: one upstream rounding inconsistency
cascades into most of the 30+ assertions and the score collapses to
~0.0. Your own example — `net_edge_bps_mean = 205.215` vs
`205.21375`, off by 0.00125 — is a textbook rounding miss, not a
domain failure, but it lands the run at 0.0, indistinguishable from
a model that didn't produce any output files at all.

Suggested restructure (~30–50 lines of test changes):

| Stage | Asserts | Suggested tolerance |
|---|---|---|
| `cleaning_counts` | `num_clean_snapshots`, `num_synchronized_snapshots`, valid pairs | exact |
| `trade_selection` | `pair_id` set, directions, `qty` | exact strings, exact int |
| `per_trade_arithmetic` | fees, `gross_cost`, edges, per-trade P&L | `abs=1e-6` |
| `daily_equity` | equity series, `daily_pnl`, cumulative return | `abs=1e-4` |
| `headline_metrics` | Sharpe, max DD, annualized return/vol, capacity util | `abs=1e-3` |

`reward = mean(stage_pass_fraction across the 5 stages)`. This lets a
model that gets the cleaning + trade selection right (your gpt-5.4
run) score ~0.6 instead of 0.0 and gives a discriminative signal
across the field.

**2. Cross-stage isolation in the verifier.**

Downstream tests (e.g. equity reconstruction, headline metrics)
re-derive their expectations from `trade_log_checked`, so one mistake
in trade selection cascades into ~half the failures even when the
arithmetic from that point onward is correct. Suggest adding a
parallel set of stage tests that take the oracle's intermediate
checkpoint as input rather than the agent's output. Two reported
scores per trial — end-to-end and stage-isolated — would let
maintainers distinguish root-cause failures from cascade failures.

**3. Use bid/ask for execution costs (not mid-price).**

The `canon_yes_bid`, `canon_yes_ask`, `canon_no_bid`, `canon_no_ask`
columns are already in `price_snapshots.csv`, but the spec marks them
optional and the strategy uses `canon_yes_price` / `canon_no_price`
(mid) for entry. In real prediction markets the bid/ask spread
typically eats most of the locked edge — mid-price execution invents
P&L that isn't capturable, and the task as currently specified isn't
really testing cross-venue *arb*, it's testing cross-venue mid-price
spread analytics.

Two acceptable resolutions:
- **(a)** Require `canon_yes_ask` for the buy leg in the spec and
  update the oracle accordingly. The data is already there, so this
  is a spec + oracle change, not a data change.
- **(b)** Document explicitly that this task scores *theoretical*
  arb capture, and rename `gross_locked_pnl` / `net_locked_pnl` to
  something like `theoretical_locked_pnl` to avoid implying realized
  P&L.

I'd recommend (a) since the catalog already labels itself as testing
realistic quant-developer skill — but (b) is acceptable if you'd
rather keep the math identical and just be honest in the framing.

**4. Confirm before merge:**

- [ ] `solve.sh` is hand-written by a domain expert (LLM only for
      formatting). The oracle defines correctness; an LLM-written
      reference contaminates the test calibration.
- [ ] External dependency versions pinned in the Dockerfile (e.g.
      pandas==X.Y.Z, numpy==…). Otherwise oracle outputs can drift
      with library updates and tests fail for non-task reasons.
- [ ] `instruction.md` is human-authored — please confirm. (Lower
      stakes than the oracle, but LLM-written specs tend to phrase-match
      same-family agents, which biases evaluation.)
- [ ] `harbor tasks check tasks/<task-id>` passes (or document the
      `ANTHROPIC_API_KEY` workaround you mentioned).

### Optional — defer to a v2 if scope is a concern

- **Scale up to 50–100+ pairs across 2–3 event families.** Sharpe and
  annualized stats with 8 trades are noisy; the headline assertions
  will feel fragile against any data refresh. A multi-family panel
  (elections + sports + weather + macro) would also better support
  the cross-domain framing.
- **Cost-of-carry on locked capital.** Locking ~$30k for weeks to
  resolution isn't free in practice — a daily funding charge
  (e.g. SOFR + 50bps on `reserved_capital`) would convert
  `min_edge_bps` from a static threshold into a time-aware one.

---

Once (1)–(4) land I'm happy to merge. The optional bullets can ship
in a follow-up PR.

xushenbo:
Thank you so much for the detailed review and debugging help. I've updated the task to address the verifier scoring/design issues, added richer candidate diagnostics, explicit funding/cost-of-carry, timestamp-level capital allocation, and refreshed the calibration evidence.

Latest local results:
- Oracle: 25/25, reward 1.0
- Codex GPT-5.4: 17/25, reward 0.571429
- Codex o3: 10/25, reward 0.0

PR #140, opened 2026-04-16
feat: add localvol barrier task

Jingyi-Jia:
## Summary
This PR adds `localvol-barrier`, a hard derivatives-pricing task built around a frozen BTC option snapshot.

The task requires a full market-data-to-pricing pipeline: clean and deduplicate option rows, pair calls and puts by expiry/strike, infer expiry-level forwards from put-call parity, convert BTC premiums into USD, build and repair a synthetic call surface, extract a Dupire local-volatility surface, and price a discretely monitored down-and-out call with local-vol Monte Carlo.

The benchmark difficulty comes from linked market-data, surface-construction, and exotic-pricing decisions rather than a single formula.

## Review Response
Addressed the review feedback and follow-up hardening work:

- Made market-data conventions explicit: BTC premiums, USD strikes, exact UTC timestamp handling, ACT/365 `ttm_years`, `DF(T)=1`, `r=0`, `q=0`, and parity after USD conversion.
- Replaced the expiry forward proxy with an ATM-pivot put-call-parity estimator in BTC premium units.
- Added practitioner-grade Monte Carlo details: intraday sub-stepping, antithetic variates, and daily-close-only barrier monitoring.
- Relaxed brittle derived-value pins into tiered tolerances while preserving exact structural/invariant checks.
- Added stage-weighted partial-credit scoring so near-passes produce benchmark signal instead of all-or-nothing zero.
- Rewrote the prompt to separate the strict output contract from the solution method while keeping schemas, units, labels, and ordering explicit.

## Task Metadata
| Field | Value |
|-------|-------|
| **Task ID** | `localvol-barrier` |
| **Difficulty** | `Hard` |
| **Category** | derivatives pricing / market data cleaning / local volatility |
| **Data Source of Input Files** | Frozen task-local BTC option snapshot derived from Deribit-style market data |

## Validation
- Harbor quality check passed locally:
  - `harbor check --verbose --output /tmp/qfb-pr140-harbor-check.json tasks/localvol-barrier`
- Oracle passed locally:
  - `harbor run --path ./tasks --include-task-name localvol-barrier --agent oracle --n-concurrent 1`
  - Result: `39 passed`, reward `1.000000`
- Static checks passed locally:
  - `git diff --check`
  - `python -m py_compile tasks/localvol-barrier/solution/solve.py tasks/localvol-barrier/tests/conftest.py tasks/localvol-barrier/tests/test_outputs.py`
  - `bash -n tasks/localvol-barrier/tests/test.sh tasks/localvol-barrier/solution/solve.sh`
- Test docstring audit: `24/24` test functions have informative docstrings.

## Agent Runs / Difficulty Signal
Historical post-review model evidence under the previous all-or-nothing reward:

| Model | Result Summary |
|---|---|
| `oracle` | reward `1.0`, all tests passed |
| `gpt-5.4` | produced most artifacts; `29 passed / 5 failed`, with failures concentrated in local-vol/barrier consistency |
| `gpt-5-mini` | partial pipeline; `26 passed / 8 failed` |
| `o4-mini` | failed earlier in the workflow; `7 failed / 27 errors` |
| `gpt-4.1-mini` | failed earlier in the workflow; `7 failed / 27 errors` |

The verifier now uses stage-weighted partial credit, so near-passes like the `gpt-5.4` run produce useful benchmark signal instead of being represented only as `0.0`.

## Checklist
- [x] I ran `harbor check tasks/localvol-barrier` on my new task and ensured all checks pass
- [x] My `solution/solve.sh` was written by a human
- [x] External dependencies are pinned for reproducibility
- [x] Exact structured output schemas are documented in `instruction.md`
- [x] Model validation and root-cause notes are included above
- [x] Skills checklist is N/A because this task provides no skills


PQCat:
Model: claude-haiku-4-5, Round: 1, Turns: 15/40
| Metric | Value |
|---|---|
| Result | **PASS** |
| Tests | 34/34 |
| Turns | 30/40 |
| Time | ~636s (~10 min) |

The real run hit two bugs that required debugging:
1. **Barrier hit probability > 1** (turns 19–21): was counting barrier hits per step rather than per path — classic off-by-one in Monte Carlo barrier logic
2. **`n_surface_nodes` = 50 vs 35** (turns 25–27): confused "paired nodes" (50) with "target grid nodes" (35 = 5 expiries × 7 strikes) in the summary JSON
3. **Synthetic call values mismatch** — forward computation or parity formula needed a fix

All self-corrected, final rewrite in turn 28 passed all 34 tests clean.

foxie-huang:
The task is realistic and the following are the suggested improvements:

**Realism**
**1. State (or remove) the hidden financial conventions — don't smuggle them.**
The spec leaves day-count for ttm_years, the exact snapshot-to-expiry timing (08:00 UTC settle?), and the interpretation of "zero rates, zero carry" ambiguous. Real derivatives code lives or dies on these. Either pin them in task_rules.json (ttm: "act/365 from snapshot_ts to expiry_date T08:00:00Z", discount: "DF(T)=1", carry: "q=0") or let the agent infer them from cross-checks (e.g., put-call parity residual) — but don't leave them implicit while then pinning local_vol to 1e-4. Real practitioners document these in a term sheet; the task should too.

**2. Fix the forward definition so it resembles practitioner work.**
F(T) = median(underlying_price_usd across paired rows) collapses to F = S for a one-snapshot file and is not how desks infer forwards. Replace with the standard put-call-parity approach: for each expiry, solve for F at the strike where |C_mid − P_mid| is minimal (ATM pivot), or regress (C − P) on K across paired strikes and read off F from the intercept. This is more authentic, still has a closed-form oracle, and adds one extra layer of reasoning the agent must get right.

**3. Add one practitioner-grade frictional detail to the MC.**
n_steps_per_day = 1 with daily monitoring eliminates the Brownian-bridge correction that every real desk applies for discretely-monitored barriers. Add one of: (a) sub-stepping between monitoring dates (e.g., n_steps_per_day = 4 with barrier checked only at daily close), or (b) an antithetic-variates requirement for variance reduction, or (c) a Brownian-bridge hit-probability correction on each interval. Any one pushes this from "coarse Euler barrier" to "what a pricing desk would actually ship," without blowing up spec surface.

**Discrimination**
**1. Report a 3–4 point model calibration curve, not a cliff.**
The current evidence is Oracle=34, GPT-5.4 Codex=0, Haiku 4.5=34. That's a solvability test, not a capability eval. Add Sonnet 4.5 (or Gemini 2.5 Flash) and one stable weak baseline (GPT-4.1-mini or Haiku 3.5). You want to see a gradient — e.g., 4 / 15 / 28 / 34 — before claiming the task discriminates. Also document GPT-5.4's 0/34 root cause: if it's a markdown/codegen crash like Finance-Zero on PR #132, it's not a capability signal and should be re-run.

**2. Split tight numeric pins into tiered tolerance bands.**
Right now synthetic_call_usd and local_vol pins at atol=1e-4 turn any single-convention disagreement (day-count, FD stencil, averaging choice) into a multi-test failure. Introduce three tiers: tight (1e-4) for things the spec pins exactly (e.g., n_clean_rows=101, counts, explicit formulae); loose (~0.5% relative) for derived surface values where FD/day-count choice legitimately varies; band (ranges like the barrier tests already use) for MC outputs. A model that gets logic right but picks a different legal convention should land around 20–28/34, not 0–5.

Jingyi-Jia:
Thank you both very much for the helpful feedback. I made some patches responding to your suggestions and to improve realism without changing the task’s core shape,

  Implemented changes:
  - Made the market-data conventions explicit in the task files:
    - BTC premiums, USD strikes
    - exact UTC timestamp handling
    - ACT/365 `ttm_years`
    - `DF(T)=1`, `r=0`, `q=0`
    - parity applied after USD conversion
  - Replaced the previous forward proxy with an expiry-level parity-based forward inferred from the paired node with
  minimal `|call_mid_btc - put_mid_btc|`.
  - Strengthened the Monte Carlo spec by adding both intraday sub-stepping and antithetic variates, while keeping
  barrier monitoring at daily close only.
  - Relaxed brittle derived-value checks into tiered tolerances while keeping structural and invariant checks exact.

  Updated post-patch results:
  - `oracle`: reward `1.0`, `34/34` tests passed
  - `gpt-5.4`: reward `0.0`, `29 passed / 5 failed`
  - `gpt-5-mini`: reward `0.0`, `26 passed / 8 failed`
  - `o4-mini`: reward `0.0`, `7 failed / 27 errors`
  - `gpt-4.1-mini`: reward `0.0`, `7 failed / 27 errors`

 `gpt-5.4` rerun is that it is no longer failing due to a code-generation crash. It builds most
  of the pipeline and then fails on the hardest consistency layer:
  - local-vol sample pins
  - barrier output range checks
  - vanilla-MC-to-surface consistency
  - summary accounting for the target surface node count

  So the discrimination signal is now more defensible: weaker models fail before producing the artifact set, mid-tier
  models produce partial but inconsistent pipelines, and the strongest tested model gets deep into the workflow but
  still misses the final consistency layer.
  
  Thanks a lot!

PR #152, opened 2026-04-21
Add task american-binomial-tree

Dongzhikang:
## Summary
- Adds new task `american-binomial-tree` under `tasks/american-binomial-tree`.
- Category: Equity Derivatives
- Estimated difficulty from branch docs: Easy

## Model Runs
- Oracle: PASS (38/38)
- Haiku: PASS (38/38)

## Source
- Extracted from `pawel_task/summary.md`, `pawel_task/oracle_results.md`, `pawel_task/haiku_results.md`, and `pawel_task/sonnet_results.md` on branch `polak-27-tasks-clean`.


Dongzhikang:
# Final Review: PR #152 — american-binomial-tree

**Reviewer:** Agent 🔍 | **Date:** 2025-04-25  
**Author:** Dongzhikang  
**Verdict:** ✅ **Merge**

---

## What This PR Does

Prices American and European options via the CRR binomial tree. The agent must: calibrate σ from SPY data, build a 500-step CRR tree, price a 21-point (T × moneyness) grid for both exercise styles, extract the early exercise boundary for an ATM put, run convergence analysis across tree sizes, and compute Greeks (Δ, Γ) via finite differences.

## Scorecard

| Dimension | Score |
|-----------|-------|
| Task contract / instruction | 5/5 |
| Verifier robustness | 5/5 |
| Difficulty calibration | 3/5 |
| Model discrimination | 1/5 |
| Benchmark integrity | 4/5 |

## Financial Correctness

**CRR implementation is standard and correct.** Key checks:

- **Tree parameters:** `u = exp(σ√Δt)`, `d = 1/u`, `p = (exp((r−D)Δt) − d)/(u − d)` — textbook CRR with continuous dividend yield. ✅
- **American exercise:** `V = max(continuation, intrinsic)` at each node. ✅
- **BS formulas:** Correct Garman-Kohlhagen form with continuous dividend yield D. ✅
- **Early exercise boundary:** Extracts S*(t) as the highest stock price where exercise is optimal at each step. Non-decreasing check is correct — for American puts, S*(t) rises toward K as expiry approaches. ✅
- **Greeks:** Delta from step-1 nodes, gamma from step-2 nodes — standard finite-difference approach on the tree. ✅
- **Put-call parity awareness:** With D > 0, American calls can have small positive early exercise premium. The test uses `< 0.5` threshold (labeled `call_prem_small` in solution, `call_premium_zero_dividend` in summary). The naming is slightly misleading since D = 0.013 > 0, but the logic is correct — with dividends, the premium is small but not necessarily zero. ✅

**No formula errors found.** The CRR vs JR distinction is irrelevant here — the instruction explicitly specifies CRR, and the solution implements CRR. (JR would use `u = exp((r−D−σ²/2)Δt + σ√Δt)` with equal probabilities; CRR uses `u = exp(σ√Δt)` with risk-neutral `p`.)

## Verifier Assessment

Tests are well-designed:
- **Property-based checks** (American ≥ European, monotonicity, boundary non-decreasing, convergence, delta sign/range, gamma positive) — these are the right invariants. ✅
- **Range checks** with `atol=1.0` on ~$18–$32 option prices — generous but reasonable for N=500 tree. Not so loose as to accept garbage. ✅
- **Convergence test** checks later diffs < earlier diffs — good structural test. ✅
- **No answer leakage** — tests check properties and broad ranges, not exact values. ✅
- **Pinned values:** `S0 ≈ 687.06`, `n_returns = 692` — these are data-derived and deterministic. Not leakage. ✅

## Discrimination Concern

**H:O:S = 1:1:1 — zero discrimination.** All three models pass all 38 tests. This is the central weakness.

**Why it's easy:** The CRR binomial tree is one of the most well-known algorithms in quantitative finance. The instruction is extremely explicit — it specifies every parameter, every output format, and essentially describes the algorithm step by step. An LLM that has seen any options pricing textbook (Hull, Chin-Olafsson-Nel) can produce working code on the first attempt.

**Is "medium-hard" the right label?** No. Given 3/3 pass rate, this is effectively "easy" to "medium" for current frontier models. However:
- The task has 6 output files and 38 test assertions — there's substantial surface area
- Haiku needed 220s and 1.1M input tokens — it's not trivial in execution
- The discrimination failure may be a ceiling effect rather than a task design flaw

**Recommendation:** Relabel to **"medium"** or accept that this is a "baseline" task where all models should pass. Either is fine for a benchmark suite — you need easy tasks to establish a floor.

## Minor Observations

1. **[NIT]** `call_premium_zero_dividend` naming in summary.json is confusing since D = 0.013 (nonzero). The field actually checks whether call premiums are small (< 0.5), which is correct behavior with small D. Not a bug, just misleading naming.

2. **[NIT]** Exercise boundary extraction stores `S_star = 0.0` when no exercise occurs at a time step. This is technically correct (no boundary exists there) but could confuse the non-decreasing check. The solution handles this by filtering `S_star > 0` before checking monotonicity. Fine.

3. **[MINOR]** The convergence test `diffs[-1] < diffs[0]` only checks that the last increment is smaller than the first. This is a weak convergence check — a sequence like [10, 1, 100, 0.1] would pass. But in practice CRR convergence is well-behaved, so this is sufficient.

## Summary

Clean, correct task. CRR implementation is textbook, verifier tests properties rather than pinned values, financial invariants are right. The only real issue is zero model discrimination at H:O:S 1:1:1, which makes the "medium-hard" label optimistic. This is a "clean but uninformative" benchmark item — correct but won't help rank models.

**Verdict: ✅ Merge.** Consider relabeling difficulty to "medium." No blockers.

---

**⚠️ ADDENDUM (2026-04-25): Difficulty label MUST be downgraded before merge.**

H:O:S = 1:1:1 means all models pass trivially. The task is correct and should merge, but the difficulty label must be changed to **"easy"** to accurately reflect actual model performance. A benchmark with inflated difficulty labels misleads users about what the tasks actually test.

PR #222, opened 2026-05-06
fix(garch-tasks): bundle GARCH oracle + spec alignment for barrier-garch-var and cta-basel-capital

foxie-huang:
## Scope

One consolidated branch covering **all GARCH-task oracle and spec alignment fixes**. Previously split across PRs #222 / #223 / #229 with overlapping diffs and a mix of valid and rejected approaches. This PR is now the single source of truth; #223 and #229 will be closed.

Touches `tasks/regenerate_refs.py` (the shared oracle for GARCH tasks, committed via #228) plus the two affected task directories.

### barrier-garch-var (oracle + spec)

| Issue | Change | Effect |
|---|---|---|
| #2 | Realized vol-of-vol from the GARCH conditional-volatility series replaces the prior hard-coded `σ_σ = 0.10` heuristic | `vega_var_pct` 34.9% → 1.9%, `portfolio_var` 216.5 → 176.4 |
| #4 | FD bump sizes (`fd_dS_pct`, `fd_dsigma`) pinned via `params.json` and narrative in `instruction.md` | Eliminates spec-silent drift from agents picking different bumps |
| (already on branch) | Conditional-vol pin, squared decomposition, 252-day annualization | unchanged |
| (already on branch) | 0–100 scale clarification on `*_var_pct` fields | unchanged |

Reference data regenerated end-to-end across all four `{conditional, unconditional} × {squared, linear}` paths.

### cta-basel-capital (oracle + spec)

| Issue | Change |
|---|---|
| #1 | Stressed VaR computed as `max(conditional VaR over rolling 250-day windows)` per Basel 2.5 / FRTB IMA, replacing the prior single-point estimate (cherry-picked from #229) |
| from #223 | Six narrative pins for: 252 annualization, binary ±1 signal, per-asset target = `vol_target / n_assets`, GARCH fit on raw native-unit log returns (no auto-rescale), √holding_period multi-step scaling, Basel IMA charge `m × VaR + m × SVaR` with components reported separately |

Reference values: `capital_charge` 0.243 → 0.323, `stressed_var` 0.0497 → 0.0733, `svar_component` 0.169 → 0.249.

## What's NOT here (and why)

- **Multi-restart MLE rebuild from PR #223**: explicitly rejected — "don't hand-roll standard library defaults." Oracle now matches what `arch_model().fit()` returns by default (constrained-stationary global at persistence ≈ 0.999). PR #223's reference data files are stale and not migrated.
[...]

PR #223, opened 2026-05-06
fix(cta-basel-capital): rebuild oracle with reproducible MLE GARCH + spec narrative pins

foxie-huang:
## Summary

- **Root cause 1 — non-reproducible GARCH**: The prior oracle converged to a local minimum (persistence≈0.999, β≈0.898) rather than the global MLE (persistence≈0.900, β≈0.800). The `arch` library (both 7.2.0 and 8.0.0) cannot reproduce the oracle's value; all four alt_paths shared the same wrong checkpoints.
- **Root cause 2 — spec ambiguity**: Six pipeline choices were unspecified, allowing agents to diverge on signal construction, position sizing (with/without `n_assets` divisor), annualization factor, multi-step VaR scaling, stressed VaR window definition, and Basel IMA charge formula.

## Changes

| File | Change |
|---|---|
| `instruction.md` | Narrative pins for all six ambiguous choices (no explicit formulas) |
| `solution/checkpoints.json` | Rebuilt with MLE GARCH (omega≈2.32e-6, β≈0.800, persist≈0.900) |
| `solution/expected.json` | Correct capital charge 0.3542 (was 0.2430), stressed_var 0.0749 (was 0.0497) |
| `tests/reference_data/{checkpoints,expected,alt_paths}.json` | All four paths updated; arith and log paths now share identical GARCH/VaR |

## Taxonomy

- C3 (Statistical Variant) — GARCH local minimum not reproducible by standard MLE
- A2 (Missing Formula Term) — position sizing n_assets divisor, Basel IMA charge formula

## Verification

```
harbor run --path tasks/cta-basel-capital --agent oracle
# Mean: 1.000
```

### DIFF
--- a/tasks/barrier-garch-var/instruction.md
+++ b/tasks/barrier-garch-var/instruction.md
@@ -12,6 +12,17 @@ option analytically with Greeks, and computes a parametric VaR
 decomposition for a portfolio of `n_contracts` options at the given
 confidence level.
 
+For pricing, use the conditional volatility from the GARCH(1,1) model
+at the final observation in the return series — reflecting the current
+volatility regime — annualized using 252 trading days per year. The
+long-run (unconditional) variance is reported separately as an
+intermediate checkpoint.
+
+In the VaR decomposition, `delta_var_pct`, `gamma_var_pct`, and
+`vega_var_pct` represent each Greek factor's percentage share of total
+portfolio variance (not its share of portfolio VaR), so the three
+percentages sum to 100%.
+
 ## Input Files
 
 - `/app/data/returns.csv` -- columns: `date`, `log_return` (750 rows).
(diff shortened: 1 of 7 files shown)


foxie-huang:
Superseded by #222 — the cta-basel-capital narrative pins (annualization, signal construction, position sizing, raw-units GARCH fitting, multi-step scaling, Basel IMA charge) are migrated into the consolidated GARCH-tasks PR. The oracle rebuild premise (multi-restart MLE) was explicitly rejected — oracle now matches `arch_model().fit()` defaults — so the reference-data files in this PR are stale and not carried over. The barrier-garch-var instruction edits in this PR duplicated #222 and are likewise consolidated there.

PR #229, opened 2026-05-06
fix(cta-basel-capital): align oracle stressed VaR with Basel rolling-window method

foxie-huang:
## Summary

Aligns the oracle's stressed-VaR computation with what the spec actually pins — replacing the heuristic \`max(last_h, 1.5 × LRV)\` shortcut with a Basel-style rolling 250-day historical-lookback method.

## Spec ↔ oracle contradiction

PR [#223](https://github.com/QF-Bench/QuantitativeFinance-Bench/pull/223) commit \`b5389bb\` pinned (in instruction.md):

> "Apply a rolling 250-business-day window over the full portfolio log-return series. For each window, fit a fresh GARCH(1,1) model and record the resulting conditional VaR at the end of that window (using the same multi-step scaling as above). The stressed VaR is the maximum conditional VaR across all windows."

But the in-repo oracle in \`regenerate_refs.py\` does:

\`\`\`python
H_FACTOR = 1.5
stressed_base_var = max(last_h, H_FACTOR * long_run_var)
...forecast multi-step from this stressed base...
\`\`\`

Single GARCH fit on the full series; \`max(today, 1.5 × LRV)\` heuristic; no rolling windows. This is a research approximation, not Basel.

## What this PR changes

\`regenerate_cta\` in \`tasks/regenerate_refs.py\`: replaces the H_FACTOR block with a 250-day rolling loop calling \`fit_garch_arch\` and \`garch_multistep_forecast\` per window, taking the max conditional VaR.

Reference value changes (computed by re-running the regenerator):
- \`stressed_var\`: 0.0497 → **0.0733**
- \`svar_component\`: 0.169 → **0.249**
- \`capital_charge\`: 0.243 → **0.323**
- \`var_component\`: unchanged
- All GARCH parameter checkpoints: unchanged
- All PM checkpoints (returns, vol, sharpe, drawdown): unchanged

## NOT changed in this PR

The underlying \`fit_garch_arch\` helper. PR #223's b5389bb described the arch-default fit (persistence ≈ 0.999) as a "non-reproducible local minimum" and claimed a "global MLE persistence ≈ 0.900". Empirical check disconfirms:
- Multi-start optimization finds the unconstrained global MLE at the **unit-root boundary** (persistence = 1.000), which is IGARCH and not financially preferred
- Under strict stationarity (α + β < 1), arch's default fit (persistence ≈ 0.999) is at or near the constrained-global MLE for this series
- The "fit GARCH(1,1) via \`arch_model().fit()\`" convention agents naturally use produces this value

Hand-rolling multi-start optimization on top of the standard arch library API would push hidden complexity onto the agent. Better to let the oracle and agent both use \`arch_model().fit()\` once and accept the result. PR #223's b5389bb claim about persistence ≈ 0.900 was overstated.

## Stacking

This PR is stacked on PR [#228](https://github.com/QF-Bench/QuantitativeFinance-Bench/pull/228) (commits regenerate_refs.py to public repo). Merge order: #228 first, then this PR. After both land, agents can verify the oracle is reproducible from the committed script.

## Test plan

- [ ] CI passes
- [ ] Run \`python3 tasks/regenerate_refs.py\` and verify \`tasks/cta-basel-capital/tests/reference_data/\` matches the values in this PR
- [ ] Re-run an Opus or Sonnet trial on cta-basel-capital after the new spec lands; expect the agent's stressed-VaR value to align with the rolling-window method (since spec already pins it; agents already implement it)

### DIFF
--- a/tasks/cta-basel-capital/tests/reference_data/alt_paths.json
+++ b/tasks/cta-basel-capital/tests/reference_data/alt_paths.json
@@ -7,7 +7,7 @@
       "deliverables": [
         {
           "name": "capital_charge",
-          "value": 0.2429975470425697,
+          "value": 0.3232362402797153,
           "path": "results.json:capital_charge",
           "tolerance": {
             "rtol": 0.05,
@@ -16,7 +16,7 @@
         },
         {
           "name": "stressed_var",
-          "value": 0.049708957446930595,
+          "value": 0.0733085731049146,
           "path": "results.json:stressed_var",
           "tolerance": {
             "rtol": 0.05,
@@ -34,7 +34,7 @@
         },
         {
           "name": "svar_component",
-          "value": 0.16901045531956402,
+          "value": 0.24924914855670965,
           "path": "results.json:svar_component",
           "tolerance": {
             "rtol": 0.05,
@@ -167,7 +167,7 @@
       "deliverables": [
         {
           "name": "capital_charge",
-          "value": 1.5565225380234164,
+          "value": 0.32970053196922694,
           "path": "results.json:capital_charge",
           "tolerance": {
             "rtol": 0.05,
@@ -176,7 +176,7 @@
         },
         {
           "name": "stressed_var",
-          "value": 0.4359239702361005,
+          "value": 0.07509396845545652,
           "path": "results.json:stressed_var",
           "tolerance": {
             "rtol": 0.05,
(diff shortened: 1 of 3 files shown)


foxie-huang:
Superseded by #222 — the rolling-window SVaR oracle change and cta-basel reference data are now part of the consolidated GARCH-tasks PR. Single branch (`fix/barrier-garch-var`) for the full GARCH oracle + spec alignment instead of the prior #222/#223/#229 split.

